PR #13, opened 2018-12-19
Add verify scripts

### LINKED ISSUE #3: Add more verify scripts
- [x] fmt
- [x] lint
  - verify-gometalinter.sh in repo-infra  is too strict, I think golint is enough
- [x] vet
- [x] boilerplate
- [x] vendor files

Reuse https://github.com/kubernetes/repo-infra when possible.

cofyc:
fixes #3


vendor/github.com/bazelbuild/buildtools/CONTRIBUTORS
```diff
@@ -0,0 +1,15 @@
+# People who have agreed to one of the CLAs and can contribute patches.
```
msau42: Is the prune setting supposed to ignore files like this?

cofyc:
Forgot to enable trigger plugin in test-infra. Now it is working.

vendor/github.com/bazelbuild/buildtools/CONTRIBUTORS
cofyc: no, these files are always be reserved, 
https://github.com/golang/dep/blob/5a7960a4cb82dda7253e8b3f921e394cd3c817d3/gps/prune.go#L157

cofyc:
`go-tests = false` is removed from k8s.io/repo-infra. We don't run tests in vendor packages.

PR #18, opened 2018-12-21
update the readme for GKE examples

kmova:
Tried to test the steps mentioned in this doc with v1.11.5-gke.5

The local ssds are mounted in the directory /mnt/disks. 

The following steps also need to be executed and not skipped to use local PVs on GKE. 
- step 2 - creation of storage class
- step 3 - creating local pv by provisioner 

Updated the text to follow through Step 2 for GKE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ requested Local SSDs. Please see
 [GKE documentation](https://cloud.google.com/kubernetes-engine/docs/concepts/local-ssd)
 for instructions for how to create a cluster with Local SSDs.
 
-Then skip to [step 4](#step-4-create-local-persistent-volume-claim).
+Then skip to [step 2](#step-2-creating-a-storageclass-19).
 
 **Note:** The raw block feature is only supported on GKE Kubernetes alpha clusters.
 


msau42:
Do the gke steps documented here, not work? https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/local-ssd#example_local_pvs

msau42:
Oh, looks like the actual link changed from what is documented in the readme here

kmova:
@msau42 - I din't actually click on the GKE link because I already had a GKE cluster with Local SSDs. My bad. 

Now I see that the GKE docs specify the creation of StorageClass as well as Deploying the Local PV provisioner. If the GKE links are updated, we cloud close this PR. Another small step that needs to be documented if trying to use example yamls in this repo with GKE:
- we need to replace the reference to the mount paths from /fast-ssds to /disks.

kmova:
@msau42 

The current flow as described in this read me for [GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/local-ssd#example_local_pvs) is to follow the steps at google cloud and then skip to [step 4](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner#step-4-create-local-persistent-volume-claim) for PVC creation. 
- The google doc creates a storage class - `local-scsi`
- This readme uses the storage class name as - `local-storage`

Now I am proposing in this PR, that flow be changed so that we follow google doc only to create a cluster with Local SSDs. Then continue with Step 2 and 3 - for Storage Class and Static Provisioner. 
- Google cloud creates the disk mounts in `/mnt/disks`
- Step3 - Local Provisioner - by [default](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/helm/provisioner/values.yaml#L47) - expects the disks to be in `/mnt/fast-disks`

May be a better approach is to close this PR and update the Google Cloud doc to:
- Update the provisioner links to new repo
- Include the steps to modify the provisioner default values from `/mnt/fast-disks` to `/mnt/disks`

kmova:
That is a good idea to update the GKE docs to be completely self contained @msau42

msau42:
I will work on updating the gke docs to be fully contained.  I'm also in the process of cleaning up the readme.  I noticed that under step 4, it already says that you should substitute the storageclassname in the PVC with whatever storage class you configured.

kmova:
Thanks @msau42. Will close this PR.

Issue #7, opened 2018-12-19
Consolidate local provisioner spec examples with helm examples


msau42:
Migrating from https://github.com/kubernetes-incubator/external-storage/issues/958
/assign

PR #26, opened 2019-01-04
Cleanup gce/gke examples

msau42:
Partially addresses #7


helm/examples/gce.yaml
```diff
@@ -1,16 +1,9 @@
 common:
-  # Beta PV.NodeAffinity field is used by default. If running against pre-1.10
-  # k8s version, the `useAlphaAPI` flag must be enabled in the configMap.
-  useAlphaAPI: false
   useNodeNameOnly: true
 classes:
 - name: local-scsi
   hostDir: "/mnt/disks/by-uuid/google-local-ssds-scsi-fs"
-  volumeMode: Filesystem
```
kmova: @msau42 - Why are these being removed?

msau42: It's filesystem mode by default, so this is redundant.  Are you seeing issues?

PR #29, opened 2019-01-04
use sigs.k8s.io which is much shorter

cofyc:
.travis.yaml which is not needed anymore is removed


cofyc:
We need update prow job first: https://github.com/kubernetes/test-infra/pull/10609

PR #34, opened 2019-01-10
Add hack/release.sh to do releasing automatically

cofyc:


cofyc:
@msau42 
This is a handy script to do releasing. 

```
docker tag v2.3.0
./hack/release.sh
```

will build and release quay.io/cofyc/local-volume-provisioner:latest and quay.io/cofyc/local-volume-provisioner:v2.3.0.

I suggest to do stable version releasing manually for now, because even if we configured a prow job to build/push image automatically, we need to check later again to make sure image are built and pushed successfully. If failed, we need to delete/create tag again to trigger prow job to rerun.

One thing I think it is useful is to automate unstable/canary image build on master branch.

What do you think?

hack/release.sh
```diff
@@ -0,0 +1,175 @@
...
+if [ -z "$SKIP_BUILD" ]; then
+    echo "info: building $image"
+    pushd provisioner &>/dev/null
+    make container REGISTRY=$REGISTRY VERSION=$VERSION
+    popd
+else
+    echo "info: building is skipped"
+fi
+
+echo "info: pushing $image"
```
msau42: Can you add a check somewhere here that if a stable image has already been pushed, we should not override it?

cofyc: Updated. https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/34/files#diff-cede583a27c227d458ca67079957c8a2R136

hack/release.sh
```diff
@@ -0,0 +1,200 @@
...
+docker_args+=(push "$image")
+docker "${docker_args[@]}"
+
+function is_latest_version() {
+    local v="$1"
+}
+
+if ! is_stable_version "$VERSION"; then
+    echo "error: VERSION '$VERSION' is not stable version, skipped pushing as latest image"
+    exit 1
```
msau42: do we want to exit with error here?  will this cause canary build jobs to fail?

cofyc: You are right. Job will fail, I will update it when configuring job in prow.

PR #38, opened 2019-01-14
Keep packages up to date

cofyc:
There is not too much we can do. Most packages have been removed in `k8s.gcr.io/debian-base-amd64`.

Some vulnerabilities exist because packages are out of date, we can fix them by upgrading them  for every new build.

### DIFF
--- a/provisioner/deployment/docker/Dockerfile
+++ b/provisioner/deployment/docker/Dockerfile
@@ -19,9 +19,14 @@ ADD . .
 RUN cd provisioner; \
 	CGO_ENABLED=0 go build -a -ldflags '-extldflags "-static"' -o local-volume-provisioner ./cmd
 
+# For security, we use kubernetes community maintained debian base image.
+# https://github.com/kubernetes/kubernetes/blob/master/build/debian-base/
 FROM k8s.gcr.io/debian-base-amd64:0.4.0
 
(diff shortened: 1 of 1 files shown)


provisioner/deployment/docker/Dockerfile
```diff
@@ -19,9 +19,14 @@ ADD . .
 RUN cd provisioner; \
 	CGO_ENABLED=0 go build -a -ldflags '-extldflags "-static"' -o local-volume-provisioner ./cmd
 
+# For security, we use kubernetes community maintained debian base image.
+# https://github.com/kubernetes/kubernetes/blob/master/build/debian-base/
 FROM k8s.gcr.io/debian-base-amd64:0.4.0
 
-RUN clean-install \
+# Keep packages up to date and install packages for our needs.
+RUN apt-get update \
```
msau42: is it possible to only update the packages when running CI builds vs doing a local build?  This may increase the container build time for local dev/testing

cofyc: docker build has builtin cache mechanism with which docker will reuse cache if nothing change in build context. In local dev/testing, docker will run all RUN commands to build image for the first time, and will only build local-volume-provisioner when user made some change in repo.

PR #37, opened 2019-01-11
Update CI links

cofyc:
fixes https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/28

PR #45, opened 2019-01-15
Save reports in ARTIFACTS directory

cofyc:
why: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/37#discussion_r247284819


cofyc:
Tested successfully, see https://gubernator.k8s.io/build/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_sig-storage-local-static-provisioner/45/pull-sig-storage-local-static-provisioner-e2e/43.

PR #46, opened 2019-01-17
use PROVISIONER_E2E_IMAGE to specify e2e image to test

cofyc:
By the way, I did some minor improvements:

- remove KUBECTL which is not necessary now
- fix bug in pass kubetest args (because I added getopts to parse cli arguments)


cofyc:
/hold
I need do some more tests

cofyc:
/hold cancel
Tested with GKE provider successfully.

Issue #50, opened 2019-01-22
A deployment issue when docker --init flag is enabled


cofyc:
We mount /dev into pod for device discovery, it has conflicts when docker --init flag is enabled because docker will mount `init` at `/dev/init`.

I reported this issue to moby project (https://github.com/moby/moby/issues/37645) and it is fixed in https://github.com/moby/moby/pull/37665. But it takes some time for this to be fixed in docker CE/EE. 

Workarounds:

- do not use docker `--init`
- do not mount `/dev`, but local volume provisioner will not discovery newly mounted devices, see https://github.com/kubernetes-incubator/external-storage/issues/783#issuecomment-395013458

It may have better workaround, but I don't have time to investigate.

cofyc:
I've added this into faq docs, close it.
https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/faqs.md#failed-to-start-when-docker---init-flag-is-enabled

PR #60, opened 2019-03-13
Mount /dev in cleanup job

cofyc:
- [x] Update k8s.io/klog to fix bug introduced in https://github.com/kubernetes/klog/pull/31
- [x] Add volume mount configuration in cleanup job Pod spec
- [x] Add e2e test

### DIFF
--- a/provisioner/pkg/deleter/jobcontroller.go
+++ b/provisioner/pkg/deleter/jobcontroller.go
@@ -253,8 +253,7 @@ func (c *jobController) RemoveJob(pvName string) (CleanupState, *time.Time, erro
 }
 
 // NewCleanupJob creates manifest for a cleaning job.
-func NewCleanupJob(pv *apiv1.PersistentVolume, volMode apiv1.PersistentVolumeMode, imageName string, nodeName string, namespace string, mountPath string,
-	config common.MountConfig) (*batch_v1.Job, error) {
+func NewCleanupJob(pv *apiv1.PersistentVolume, volMode apiv1.PersistentVolumeMode, imageName string, nodeName string, namespace string, mountPath string, config common.MountConfig) (*batch_v1.Job, error) {
 	priv := true
 	// Container definition
 	jobContainer := apiv1.Container{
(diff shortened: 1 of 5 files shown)


cofyc:
/assign @msau42 

1) `should discover dynamically created local persistent volume mountpoint in discovery directory` is removed, because it gets exercised in `should create and recreate local persistent volume` already
2) `should create and recreate local persistent volume` is tested in a matrix to cover more cases

List of current e2e tests:

- [sig-storage] PersistentVolumes-local  Local volume provisioner [Serial] should not create local persistent volume for filesystem volume that was not bind mounted
- [sig-storage] PersistentVolumes-local  Local volume provisioner [Serial][UseJobForCleaning: true [VolumeType: dir] should create and recreate local persistent volume
- [sig-storage] PersistentVolumes-local  Local volume provisioner [Serial][UseJobForCleaning: false][VolumeType: dir] should create and recreate local persistent volume
- [sig-storage] PersistentVolumes-local  Local volume provisioner [Serial][UseJobForCleaning: true][VolumeType: block] should create and recreate local persistent volume
- [sig-storage] PersistentVolumes-local  Local volume provisioner [Serial][UseJobForCleaning: false][VolumeType: block] should create and recreate local persistent volume
- [sig-storage] PersistentVolumes-local  Stress with local volume provisioner [Serial] should use be able to process many pods and reuse local volumes

provisioner/pkg/deleter/jobcontroller.go
```diff
@@ -290,7 +289,21 @@ func NewCleanupJob(pv *apiv1.PersistentVolume, volMode apiv1.PersistentVolumeMod
...
+			Name: "dev",
+			VolumeSource: apiv1.VolumeSource{
+				HostPath: &apiv1.HostPathVolumeSource{
+					Path: "/dev",
+				},
+			},
+		})
+		jobContainer.VolumeMounts = append(jobContainer.VolumeMounts, apiv1.VolumeMount{
+			Name:      "dev",
+			MountPath: "/dev",
```
msau42: I'm assuming this doesn't have any issues if we override the container's /dev?

cofyc: No, to discover and cleanup we need to mount host `/dev` into provisioner and cleanup containers.
Potential issue is in [certain docker setup](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/50) program cannot start when host `/dev` is mounted into container, but provisioner cannot too.

msau42: Can you add a comment with a link to the issue?

I assume mounting /dev somewhere else won't work because of symlink resolution issues?

cofyc: Updated.

> I assume mounting /dev somewhere else won't work because of symlink resolution issues?
Yes

PR #61, opened 2019-03-13
Use k8s.io/klog

cofyc:
- deprecate github.com/golang/glog in favor of k8s.io/klog
- use latest sigs.k8s.io/sig-storage-lib-external-provisione which uses k8s.io/klog


Gopkg.toml
```diff
@@ -11,8 +11,8 @@ required = [
 ]
 
 [[constraint]]
-  name = "github.com/kubernetes-sigs/sig-storage-lib-external-provisioner"
-  version = "1.0.0"
+  name = "sigs.k8s.io/sig-storage-lib-external-provisioner"
+  branch = "master"
```
msau42: do we want master or the 3.0.0-beta?

Gopkg.toml
```diff
@@ -11,8 +11,8 @@ required = [
...
 
 [[constraint]]
   name = "k8s.io/client-go"
```
msau42: i'm actually wondering if we need all of these constraints? Do we depend on any API changes that are not officially tagged?

provisioner/cmd/main.go
```diff
@@ -45,6 +45,7 @@ var (
 
 func main() {
 	rand.Seed(time.Now().UTC().UnixNano())
+	klog.InitFlags(nil)
```
msau42: I remember there was some strange ordering with klog and flags that we had to deal with with csi sidecars.  See an [example](https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L76)

Gopkg.toml
cofyc: We do not depend on any API changes that are not officially tagged. If we need, we can wait until they are released.

We need these constraints, all k8s.io packages must be same version (at lease same release version), because they are changed together in k/k staging directory.

Gopkg.toml
cofyc: I've removed this constraint section because it's a standalone package and we can review `Gopkg.lock` to see if it has been changed by accident.

If we want to use newer version we can update it to laster master by using `dep ensure -update` or a specify version by using `dep ensure -add sigs.k8s.io/sig-storage-lib-external-provisioner@<version>`.

provisioner/cmd/main.go
cofyc: The ordering is simple:

- register CLI flags on flagset (we use go default flagset here)
- parse flags
- override flags (e.g. enable `logtostderr` always)

Your example is complex because it uses `pflag` to parse CLI flags, so it must register go standard flagset onto `pflag` flagset and call `Parse` on `pflag`.

Gopkg.toml
msau42: client-go and api need to be in sync because they're API libraries, but do all the others?

If you remove the constraints and do an `dep ensure -update` do the versions it pulls in look ok?  By default, dep is supposed to pull in the latest official release tag, otherwise master.

cofyc: > client-go and api need to be in sync because they're API libraries, but do all the others?

It is better to be in sync, because they are kept in sync with k/k code in [staging directory](https://github.com/kubernetes/kubernetes/tree/master/staging), e.g. 1.14 k/k code probably cannot compile with 1.13 k8s.io/apimachinery.

> If you remove the constraints and do an dep ensure -update do the versions it pulls in look ok? By default, dep is supposed to pull in the latest official release tag, otherwise master.
yes

Another reason is sometimes we need to use a specific version of dependency (from k8s.io package's Godeps.json), e.g. https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/2d414955ebad92b5477a573fc835ca31a88fe701/Gopkg.toml#L45-L48. It's best to constrain versions of them and upgrade until needed.

msau42: Ok I guess that is because we import packages from k/k directly instead of only staging libraries which are supposed to be independent from each other.

If we can eventually move away from k/k dependencies then I think we shouldn't need these constraints.  There's already work to move the mount libraries in k/k into staging, so that will help with one dependency.  After mount is moved out, then maybe part of fs util could move out too.  And the kubelet/api dependency I think is easy to get rid of.

PR #63, opened 2019-03-16
Fix issues with priorityClassName field

### LINKED ISSUE #62: Should add conditional of priorityClassName
so priorityClassname can only be run in the kube-system namespace. if they specify any other namespace it fails. The template should check for a conditional, or straight up remove it (which I think would be best, considering its on the chopping block for .14 and beyond). 

the offending line is here
https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/helm/provisioner/templates/provisioner.yaml#L67

cofyc:
fixes #62
- [x] disable it by default, because we deploy provisioner in default namespace by default
- [x] ignore non system critical classes ("system-cluster-critical" "system-node-critical") in non kube-system namespaces (see [hard coded code in priority admission plugin](https://github.com/kubernetes/kubernetes/blob/df2094b3d728bd58c5f23b41add109e32fc7c301/plugin/pkg/admission/priority/admission.go#L143-L155))
- [x] add an example to demonstrate how to [make provisioner critical in cluster](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical)


helm/provisioner/templates/provisioner.yaml
```diff
@@ -64,7 +64,16 @@ spec:
         app: local-volume-provisioner
     spec:
       serviceAccountName: {{.Values.daemonset.serviceAccount}}
+{{- if .Values.daemonset.priorityClassName }}
+{{- if eq .Values.common.namespace "kube-system" }}
       priorityClassName: {{.Values.daemonset.priorityClassName}}
+{{- else }}
+      {{- $systemCriticalClasses := list "system-cluster-critical" "system-node-critical" }}
+      {{- if not (has .Values.daemonset.priorityClassName $systemCriticalClasses) }}
+      priorityClassName: {{.Values.daemonset.priorityClassName}}
```
msau42: should we have a test case where you can set a non-critical priority class?

cofyc: Updated, two examples are added in this PR:

- helm/examples/baremetal-priority-critical.yaml
- helm/examples/baremetal-priority-noncritical.yaml

Generated manifest files are tested manually. (try to test against cluster automatically in #64)

msau42:
/lgtm
Thanks!

PR #69, opened 2019-03-27
Update discovery.go

segator:
New PV's have the format <foldername>-<hash>

### DIFF
--- a/provisioner/pkg/discovery/discovery.go
+++ b/provisioner/pkg/discovery/discovery.go
@@ -289,7 +289,7 @@ func generatePVName(file, node, class string) string {
 	h.Write([]byte(node))
 	h.Write([]byte(class))
 	// This is the FNV-1a 32-bit hash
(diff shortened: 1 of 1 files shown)


msau42:
Thanks for the PR! Can you describe exactly what issue you're seeing with the current naming scheme?

MOZGIII:
I can. With the current naming scheme there's no correspondence between PV and the backing directory / block symlink. What this PR proposes is to provide that correspondence.

I feel the pain that's causing this, however I'd go a different route. Changing the naming scheme would work, but adding a label that would provide a bare file name would work even better! It would allow me to map the PVs to the actual on-disk folders, which, in turn, would allow me to bind PVC to the exact PV I want to bind. Currently I have to tweak volume sizes in such a way that k8s picks the right one - but this is hell.

MOZGIII:
I've created #133

PR #70, opened 2019-03-28
should not swallow error when listing mounter

tcmichael:


cofyc:
Using `mountPointsErr` here does not have any problem because `discoverVolumesAtPath` does not return error, but it's good to use same variable name.

/lgtm
/approve
Thanks! By the way, you need to follow instructions to sign the CLA.

tcmichael:
I signed it

PR #84, opened 2019-04-23
More faqs

cofyc:
- Why it's best to use UUID of disk in mount point path
- Why I need to bind mount normal directories to create PVs for them
- Another case when PV with delete reclaimPolicy is released but not
going to be reclaimed

### DIFF
--- a/docs/faqs.md
+++ b/docs/faqs.md
@@ -2,6 +2,8 @@
 
 ## Table of Contents
 
+- [Why it's best to use UUID of disk in mount point path](#why-its-best-to-use-uuid-of-disk-in-mount-point-path)
+- [Why I need to bind mount normal directories to create PVs for them](#why-i-need-to-bind-mount-normal-directories-to-create-pvs-for-them)
 - [I updated provisioner configuration but volumes are not discovered](#i-updated-provisioner-configuration-but-volumes-are-not-discovered)
 - [I bind mounted a directory into sub-directory of discovery directory, but no PVs created](#i-bind-mounted-a-directory-into-sub-directory-of-discovery-directory-but-no-pvs-created)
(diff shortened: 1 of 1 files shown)


docs/faqs.md
```diff
@@ -91,6 +112,19 @@ Of course, on a specific platform if you have a reliable mechanism to detect if
...
+If by accident kubernetes node object has been recreated, new node object with
+same name for same node will get a new UID. By default, provisioner includes
+node.UID in its name, it will filter out local PVs that were not created with
+this name and do not reclaim these PVs. You can check
+`pv.kubernetes.io/provisioned-by` annotation in PV object.
+
+You can use enable `useNodeNameOnly` to set provisioner name without node UID
+to solve this cleanup issue. However, this will only fix new deployments going
+forward. Existing PVs created with an older local provisioner will still have
+the UID provisioner naming.
```
cofyc: https://github.com/kubernetes-incubator/external-storage/issues/945#issuecomment-414393266
Can we enable useNodeNameOnly by default in next release?

docs/faqs.md
```diff
@@ -10,6 +12,34 @@
...
+local volumes](operations.md#deletingremoving-the-underlying-volume).
+
+## Why I need to bind mount normal directories to create PVs for them
+
+This is because there is a race between mounting another filesystem volume on a
+normal directory and creating a PV for it. If you want to create PVs for normal
+directories which do not have a mount point, you need to bind mount them onto
+another directory under discovery directory, or themselves if they are already
+in. Mount point on directory explicitly express it is ready to have a PV
+created for it.
```
cofyc: https://github.com/kubernetes-incubator/external-storage/issues/370

docs/faqs.md
msau42: We can enable it by default but probably need to cut a new major version

PR #88, opened 2019-04-24
Unset DOCKER_CONFIG in hack/run-e2e.sh

cofyc:
Unset DOCKER_CONFIG  because we don't need it and it may be read-only which will cause the script to fail


cofyc:
/assign @msau42
however, I'm not sure, is there a way to trigger post job in test-infra to run except pushing the tag again?

PR #90, opened 2019-04-25
Filter out unstable tags

cofyc:


cofyc:
/assign @msau42 

`hack/release.sh` failed to push `latest` image because of this

PR #92, opened 2019-04-28
Update k8s code to 1.14 and use general host executor

### LINKED ISSUE #59: Use general host executor
Why: simpler
How: https://github.com/kubernetes/kubernetes/pull/74391

cofyc:
- 0397c7d7455443571acc3a6f2fbd44adb5e13372 vendor files
- 2eba6f1dcbe67229f124d3b0c5ea3b310c4d3958 Update k8s code to 1.14 and use general host executor

fixes #59


Gopkg.toml
```diff
@@ -12,33 +12,37 @@ required = [
...
 
 [[constraint]]
   name = "k8s.io/api"
-  branch = "release-1.13"
+  branch = "release-1.14"
 
 [[constraint]]
   name = "k8s.io/kubernetes"
-  branch = "release-1.13"
+  branch = "release-1.14"
```
msau42: we could use the kubernetes-1.14.0 tags instead

cofyc: done

PR #73, opened 2019-04-08
Support for PodSecurityPolicies

weargoggles:
Add a PodSecurityPolicy, Role and RoleBinding to allow the provisioner's security contexts on clusters where PodSecurityPolicy admission control is in effect.

Fixes #72

Issue #64, opened 2019-03-16
Add tests for helm examples


cofyc:
Why: #62, https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/93

Test cases:

- [ ] pod security policy when PodSecurityPolicy admission control is in effect

PR #93, opened 2019-04-29
Make pod security policy in Helm chart valid

weargoggles:
The earlier PR #73 contained an invalid PodSecurityPolicy. Oops.


msau42:
/assign @cofyc 
are we able to have an e2e test config that uses pod security policy?

cofyc:
> are we able to have an e2e test config that uses pod security policy? 

I've opened an issue to test helm example (https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/64), but this is not easy to test because it requires PodSecurityPolicy admission plugin is in effect.

PR #96, opened 2019-05-06
Support cross-building and multi-arch image

### LINKED ISSUE #95: local-volume-provisioner for arm
Hi,
I need to configure the local-storage-provisioner for an arm cluster, but the `quay.io/external_storage/local-volume-provisioner` image does not work on arm devices. 
Is it possible to do that? And, if it is, how? Can I get the Dockerfile to change the base image or exists any image for that? Thanks.

cofyc:
fixes #95 

## Overview

This is the same as Kubernetes official image release process.

First, we push image for each architecture to `local-volume-provisioner-<arch>:<tag>` repo separately.

- local-volume-provisioner-amd64
- local-volume-provisioner-arm
- local-volume-provisioner-arm64
- local-volume-provisioner-ppc64le
- local-volume-provisioner-s390x

Second, we push a multiple-arch manifest to the main repository `local-volume-provisioner`.

When users run `docker pull local-volume-provisioner`, the docker engine does the work of selecting which image to pull based on the operating system and architecture on which it is running.

## Supported arches

`amd64 arm arm64 ppc64le s390x`

These are what Kubernetes official images support.

### References

- https://github.com/docker-library/official-images#multiple-architectures
- https://docs.docker.com/engine/reference/commandline/manifest/
- https://github.com/estesp/manifest-tool
- https://console.cloud.google.com/gcr/images/google-containers/GLOBAL/debian-base
- https://github.com/kubernetes/kubernetes/tree/master/build/debian-base
- https://github.com/prometheus/busybox

### DIFF
--- a/third_party/multiarch/qemu-user-static/register/qemu-binfmt-conf.sh
+++ b/third_party/multiarch/qemu-user-static/register/qemu-binfmt-conf.sh
@@ -1,0 +1,367 @@
+#!/bin/sh
+# enable automatic i386/ARM/M68K/MIPS/SPARC/PPC/s390/HPPA/Xtensa/microblaze
+# program execution by the kernel
+#
+# downloaded from https://raw.githubusercontent.com/qemu/qemu/master/scripts/qemu-binfmt-conf.sh
+
+qemu_target_list="i386 i486 alpha arm armeb sparc32plus ppc ppc64 ppc64le m68k \
+mips mipsel mipsn32 mipsn32el mips64 mips64el \
+sh4 sh4eb s390x aarch64 aarch64_be hppa riscv32 riscv64 xtensa xtensaeb microblaze microblazeel"
+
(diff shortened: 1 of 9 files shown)


cofyc:
/hold
quay.io does not support manifest by default, I've contacted quay.io to enable for local-volume-provisioner and local-volume-provisioner-{s390x,ppc64le,arm,arm64,amd64}.

cofyc:
/assign @msau42 
What do you think?

third_party/multiarch/qemu-user-static/register/qemu-binfmt-conf.sh
```diff
@@ -0,0 +1,367 @@
+#!/bin/sh
```
cofyc: `third_party/multiarch/qemu-user-static` is required because we need to run `apt` binary to update packages in debian-base image.

msau42:
/lgtm
Thanks for putting this all together!

cofyc:
manifests support has been enabled on our repositories.

Issue #99, opened 2019-05-14
better to have only one StorageClass name


silenceshell:
There are (at least) three StorageClass names for local volume:

- [fast-disks](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/eab7af7b2497b8384733441ec88a088f2653a041/deployment/kubernetes/example/default_example_storageclass.yaml#L5)
- [local-fast](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/getting-started.md#option-1-using-the-local-volume-static-provisioner)
- [local-storage](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/getting-started.md#step-4-create-local-persistent-volume-claim)

Is it better to have only one StorageClass name? If so, which is better?
I am glad to submit a PR to fix this.

cofyc:
It's ok to configure multiple storage classes with any valid names for local PVs in Kubernetes. However, in one place, it's better to use one name. Could you change `local-fast` to `local-storage` in [getting started docs](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/getting-started.md)?

msau42:
I think it's fine to change fast-disks to local-storage too, so that our examples work out of the box with the default helm deployment.

silenceshell:
@cofyc @msau42 Thanks for your response! I will submit a PR later.

PR #101, opened 2019-05-16
rename StorageClass name from local-fast to local-storage

silenceshell:
fix issue #99


msau42:
/ok-to-test
/lgtm
/approve

Thank you!

PR #102, opened 2019-05-17
add retry for getting node when setup

yuyulei:
when machine reboot, local-volume-provisioner started as daemonset, try to get node info with ClusterIP of svc:kubernetes. 

But at the same time, kube-proxy maybe not start, so the iptable rule(ClusterIP of svc:kubernetes -> apiserver address) has not been created/updated. 

So find this :

`F0515 11:48:50.664856       1 main.go:119] Could not get node information: Get https://<ClusterIP of  svc:kubernetes>:443/api/v1/nodes/<node_name>: dial tcp <ClusterIP>:443: connect: connection timed out`

After kube-proxy started, local-volume-provisioner will start successfully. 

Add retry can protect local-volume-provisioner from crashing.

### DIFF
--- a/cmd/local-volume-provisioner/main.go
+++ b/cmd/local-volume-provisioner/main.go
@@ -38,6 +38,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const maxGetNodesRetries = 3
+
 var (
 	optListenAddress string
 	optMetricsPath   string
@@ -111,9 +113,20 @@ func main() {
 }
 
 func getNode(client *kubernetes.Clientset, name string) *v1.Node {
-	node, err := client.CoreV1().Nodes().Get(name, metav1.GetOptions{})
-	if err != nil {
-		klog.Fatalf("Could not get node information: %v", err)
+	var retries int
+
+	for {
+		node, err := client.CoreV1().Nodes().Get(name, metav1.GetOptions{})
+		if err == nil {
+			return node
+		}
+
+		retries++
+		klog.Infof("Could not get node information (remaining retries: %d): %v", maxGetNodesRetries-retries, err)
+
+		if retries >= maxGetNodesRetries {
+			klog.Fatalf("Could not get node information: %v", err)
+		}
+		time.Sleep(time.Second)
 	}
-	return node
 }


cofyc:
hi, `local-volume-provisioner` is a long running program, you should set `restartPolicy` to `Always`.  The kubelet will restart it if it fails to get the node. It's best practice to fail on fatal errors in production and let process supervisor restart the process.

yuyulei:
In actual, this crash is stable probability to reproduce. 

Esay retries can help to avoid an expected crash. Now the provisioner is stateless, so it works by restarting. 

In the actual production environment, there maybe are some monitors watching all kinds failure of pods and nodes. Assuming that reboot happens, the monitor will report the crash of provisioner possibly, but we no need to handle this. The monitor just confuse us .

cmd/local-volume-provisioner/main.go
```diff
@@ -111,9 +113,17 @@ func main() {
...
-		klog.Fatalf("Could not get node information: %v", err)
+	var retries int
+
+	for {
+		node, err := client.CoreV1().Nodes().Get(name, metav1.GetOptions{})
+		if err == nil {
+			return node
+		}
+
+		retries++
```
cofyc: add a log here and one-second sleep, then the user knows what happens, e.g.

```
klog.Infof("Could not get node information (remaining retries: %d): %v", maxGetNodesRetries - retries, err)
time.Sleep(time.Second)
```

cmd/local-volume-provisioner/main.go
```diff
@@ -111,9 +113,17 @@ func main() {
...
+		if retries == MaxGetNodesRetries {
```
cofyc: `==` => `>=`

cmd/local-volume-provisioner/main.go
```diff
@@ -38,6 +38,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const MaxGetNodesRetries = 3
+
```
cofyc: `MaxGetNodesRetries` -> `maxGetNodesRetries`

cmd/local-volume-provisioner/main.go
cofyc: a small sleep ensures that the provisioner will not fail in a short time if we want it to tolerate network failures.

cofyc:
/lgtm

hi, @yuyulei Thanks! 

By the way, you need to sign the CLA. You can follow instructions [here](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/102#issuecomment-493316165).

yuyulei:
I signed it

cofyc:
/lgtm
/approve
Thanks for your contribution!

Issue #116, opened 2019-06-20
Allow additional labels to be attached to PVs provisioned by the provisioner


gnufied:
Ideally, I would like to accurately track PVs that were provisioned by this provisioner so as it can better integrate with higher level tools like - https://github.com/openshift/local-storage-operator

My own use case is:

1. Use creates a CR that triggers DaemonSet managed by provisioner and local-storage-operator.
2. Provisioner daemon provisions PVs as expected.
3. User deletes the CR. At this point, I do not accurately know if it is safe to delete the LocalVolume CR or if there are PVs that are in use and deleting the CR will also delete the daemonsets managed by it and hence those in-use PVs will not be properly cleaned up (may have left over data).

cc @msau42 @cofyc

cofyc:
Yes, I'd like to allow additional labels to be attached to provisioned PVs (xref: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/52#issuecomment-484014519).

For your use case, does `pv.kubernetes.io/provisioned-by` annotation serve your purpose?

gnufied:
@cofyc The `provisioned-by` annotation includes node name(and some random hash?) IIRC. For example:

```
pv.kubernetes.io/provisioned-by: local-volume-provisioner-127.0.0.1-1ab94f1b-ba0b-42db-a0b1-98890017377f
```

^ which won't work for my use case. The other thing obviously is - user may have been directly using local-storage-provisioner whereas I am interested in tracking PVs that are were created by the `LocalVolume` CR.

So yeah I think ability to specify custom labels in configmap should work for my use case. I don't particularly need per-storageclass-configuration labels but same set of labels for all PVs is what I am looking for. I will see if I can open a PR for this.

Issue #52, opened 2019-01-31
PersistentVolume naming


predakanga:
To simplify administration, it would be helpful to have a friendly name embedded in the generated PV's name.

This could be accomplished by adding the mount point/device name before the checksum.

For example, the following PV:
```Name:              local-pv-27f18fc5
Labels:            <none>
Annotations:       pv.kubernetes.io/provisioned-by: local-volume-provisioner-dev-b7cb269c-21f6-11e9-8d64-000c291c888b
Finalizers:        [kubernetes.io/pv-protection]
StorageClass:      hdd
Status:            Available
Claim:
Reclaim Policy:    Retain
Access Modes:      RWO
VolumeMode:        Filesystem
Capacity:          19Gi
Node Affinity:
  Required Terms:
    Term 0:        kubernetes.io/hostname in [dev]
Message:
Source:
    Type:  LocalVolume (a persistent volume backed by local storage on a node)
    Path:  /mnt/disks/gitlab
Events:    <none>
```

would become `local-pv-gitlab-27f18fc5`

This shouldn't cause any problems, as the name is already used to generate the checksum - any changes in filename would already cause a change in PV name.

cofyc:
I agree current naming is not friendly for managing PVs. However, PV name is currently used as a unique identifier for the path on the node. And the more flexible way to help PV administration is to add labels, e.g.

```
- local.storage.k8s.io/hostname = <node-name>
- local.storage.k8s.io/storageclass = <storage-class-name>
```

Additionally, provisioner configuration can support extra labels for discovered PVs for each storage class, e.g.

```
local-storage-gitlab:
  hostDir: /mnt/disks/gitlab
  mountDir: /mnt/disks/gitlab
  labels:
     app=gitlab
local-storage-other:
  hostDir: /mnt/disks/other
  mountDir: /mnt/disks/other
  labels:
     app=other
```

Then, you can run `kubectl get pv -l app=gitlab` to list all gitlab PVs.

What do you think? cc @msau42

msau42:
I like the labels idea. We also originally went with the hashing approach because we didn't want to potentially leak sensitive information into the PV name which may be readable globally

KevinTHU:
now the node information of pv can only be shown in the **annotation** or **Node Affinity**, I don't think it is a friendly way (at least the info could be searched, like label, or the best directly on pv name), hope for consideration.

cofyc:
[...]

PR #118, opened 2019-06-21
Allow user to configure additional PV labels

gnufied:
Allow users to configure additional PV labels.

Fixes https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/116

### DIFF
--- a/cmd/local-volume-provisioner/main.go
+++ b/cmd/local-volume-provisioner/main.go
@@ -94,6 +94,7 @@ func main() {
 		UseNodeNameOnly:   provisionerConfig.UseNodeNameOnly,
 		Namespace:         namespace,
 		JobContainerImage: jobImage,
+		LabelsForPV:       provisionerConfig.LabelsForPV,
 	})
 
 	klog.Infof("Starting metrics server at %s\n", optListenAddress)
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -105,6 +105,8 @@ type UserConfig struct {
 	// UseNodeNameOnly indicates if Node.Name should be used in the provisioner name
 	// instead of Node.UID.
 	UseNodeNameOnly bool
+	// LabelsForPV stores additional labels added to provisioned PVs
+	LabelsForPV map[string]string
 }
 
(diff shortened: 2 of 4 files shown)


cofyc:
hi, IIUC, in [local-storage-operator](https://github.com/openshift/local-storage-operator), the user may create multiple `LocalVolume` CRs. Each `LocalVolume` CR will create local-volume-provisioner instances on selected nodes to discovery local volumes. Is it possible to create multiple instances of local-volume-provisioner on the same node? If this is true, `pv.kubernetes.io/provisioned-by` cannot be used in your case because [it is generated with `node.Name` and `node.UID`](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/f281c24b6b3efc381989dfabb2e9c8bac7e97c78/pkg/controller/controller.go#L47-L52).

How about supporting labels per storage class instead? It can serve two purposes:

- in local-storage-operator, you can add labels for every storage class to manage all PVs
- users can [attach custom labels for each storage class](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/52#issuecomment-484014519)

gnufied:
@cofyc do we really need per-storageclass labels? It seems to me that - PVs already have StorageClass associated with them, so one can easily isolate PVs provisioned by local-storage-provisioner given combination of PV labels and storageclasses. 

For example - one can find all PVs labeled with labels proposed in this PR and then one can group the PVs by storageclass by searching for matching StorageClass. Do you think we really need per-storageclass label?

cofyc:
e2e fails because of [this](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/120).

cofyc:
/approve

I agree not to support extra labels per storage class. I'm going to add a built-in label for storage class with which users can list PVs by this built-in label https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/121.

cofyc:
Can you support this configuration in helm chart? https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/tree/master/helm/provisioner

PR #121, opened 2019-06-26
Add builtin storage class label

### LINKED ISSUE #52: PersistentVolume naming

cofyc:
fixes #52

This is useful to list PVs by storage class name, e.g. 

```
kubectl get pv -l local.storage.k8s.io/storageclass # all local pvs provisioned by provisioner
kubectl get pv -l local.storage.k8s.io/storageclass=local-ssd # all local pvs for storage class local-ssd
kubectl get pv -l local.storage.k8s.io/storageclass=local-hdd # all local pvs for storage class local-hdd
```

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -27,7 +27,8 @@ import (
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/metrics"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
 	storagev1listers "k8s.io/client-go/listers/storage/v1"
 	"k8s.io/client-go/tools/cache"
 	esUtil "sigs.k8s.io/sig-storage-lib-external-provisioner/util"
@@ -304,6 +305,17 @@ func (d *Discoverer) createPV(file, class string, reclaimPolicy v1.PersistentVol
 	klog.Infof("Found new volume at host path %q with capacity %d, creating Local PV %q, required volumeMode %q",
 		outsidePath, capacityByte, pvName, volMode)
 
+	pvLabels := make(map[string]string)
+	for k, v := range d.Labels {
+		pvLabels[k] = v
+	}
+	// truncate the class name if it exceeds the max allowed length
+	if len(class) > validation.LabelValueMaxLength {
+		pvLabels[common.LocalStorageClassLabel] = class[:validation.LabelValueMaxLength]
+	} else {
+		pvLabels[common.LocalStorageClassLabel] = class
+	}
+
 	localPVConfig := &common.LocalPVConfig{
 		Name:            pvName,
 		HostPath:        outsidePath,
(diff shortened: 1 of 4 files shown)


pkg/discovery/discovery.go
```diff
@@ -304,6 +305,17 @@ func (d *Discoverer) createPV(file, class string, reclaimPolicy v1.PersistentVol
...
+	pvLabels := make(map[string]string)
+	for k, v := range d.Labels {
+		pvLabels[k] = v
+	}
+	// truncate the class name if it exceeds the max allowed length
+	if len(class) > validation.LabelValueMaxLength {
+		pvLabels[common.LocalStorageClassLabel] = class[:validation.LabelValueMaxLength]
+	} else {
+		pvLabels[common.LocalStorageClassLabel] = class
+	}
```
cofyc: The PV will be rejected by the API server with a long label value (> 63 chars). Because the storage class name is for information only, it's safe to truncate it.

gnufied: Thats a good point. I think I should have done this for user configured labels that I added in my previous PR. Can we fix that here too or do you want me to open a follow up PR?

msau42:
I'm unclear if we really need to add a label for storageclass, since it's already an attribute in the PV object.  Can you accomplish something similar by using the -json flag to get the storageclass fields and filter pv objects based on that?

cofyc:
Yes, it's possible but complex.

```
$ kubectl get pv -ojsonpath="{range .items[?(@.spec.storageClassName=='local-storage')]}{.metadata.name}{'\n'}{end}"
local-pv-8f263585
local-pv-d6e1bd86
local-pv-da99535d
local-pv-e53dce3
```

cofyc:
/close

Closing this because there are some alternatives.

```
$ kubectl get pv  | grep 'local-storage'
local-pv-8f263585   18Gi       RWO            Delete           Available           local-storage            35m
local-pv-d6e1bd86   10Mi       RWO            Delete           Available           local-storage            37m
local-pv-da99535d   10Mi       RWO            Delete           Available           local-storage            37m
local-pv-e53dce3    10Mi       RWO            Delete           Available           local-storage            37m
```

pkg/discovery/discovery.go
cofyc: I don't think it is necessary for user-configured labels. Users should fix their labels if the local-volume-provisioner cannot create PVs, otherwise, they may fail to list PVs with their PV labels.

PR #123, opened 2019-07-01
Set PV owner reference

h0tbird:
When a `PV` is created it is set to be dependent of the `Node` backing it.
When the `Node` is deleted the `PV` is also deleted by the k8s garbage collector.

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -31,7 +31,7 @@ import (
 
 	"hash/fnv"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
@@ -107,6 +107,8 @@ type UserConfig struct {
 	UseNodeNameOnly bool
 	// LabelsForPV stores additional labels added to provisioned PVs
 	LabelsForPV map[string]string
+	// SetPVOwnerRef indicates if PVs should be dependents of the owner Node
+	SetPVOwnerRef bool
 }
 
 // MountConfig stores a configuration for discoverying a specific storageclass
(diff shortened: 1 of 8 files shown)


msau42:
/ok-to-test

We should consider if there's any use cases for leaving the PVs around, and would any users be depending on this current behavior.  If there are such use cases, then this may need to enabled as a new option to not break backwards compatibility.

cc @cofyc @gnufied

gnufied:
I can't think of a reason of not doing this. so +1 from me.

cofyc:
add an option for this?

cofyc:
LGTM

@msau42 What do you think?

pkg/common/common.go
```diff
@@ -199,6 +203,8 @@ type ProvisionerConfiguration struct {
 	// LabelsForPV could be used to specify additional labels that will be
 	// added to PVs created by static provisioner.
 	LabelsForPV map[string]string `json:"labelsForPV" yaml:"labelsForPV"`
+	// SetPVOwnerRef indicates if PVs should be dependents of the owner Node, default to false
+	SetPVOwnerRef bool `json:"setPVOwnerRef" yaml:"setPVOwnerRef"`
```
msau42: Can we add this option to the helm chart (default off) too?

h0tbird: Sure, done here: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/123/commits/74869f0c8ecc273c543618c1d37892aebb1dfbd3

h0tbird:
The test is failing with the error:
```
The command '/bin/sh -c apt-get update     && apt-get upgrade -y     && clean-install     util-linux     e2fsprogs     bash' returned a non-zero code: 1
```
Not much I can do about it, I will retry later.

cofyc:
Thanks! LGTM. Can you squash the commits a bit?

cofyc:
Thanks!
/lgtm
/approve

Issue #6, opened 2018-12-19
Delete local PV If directory not found


msau42:
Migrating issue from: https://github.com/kubernetes-incubator/external-storage/issues/1052

/help

cofyc:
Thinking more about this (previous [thoughts](https://github.com/kubernetes-incubator/external-storage/issues/1052#issuecomment-436864718)). It's hard to see delete local PV if directory not found is a good idea.

Possible scenarios when the volume is not found on deletion:

- node gets recreated with the same name: https://github.com/kubernetes-incubator/external-storage/issues/1052#issue-378532378
- node reboots but didn't mount volumes (because of invalid `/etc/fstab` file or something)

If we deleted PVs in these cases in provisioner, users are in charge of cleaning up volumes themselves which is a burden, otherwise, data in volumes may be leaked to other programs in the future.

Instead of letting users clean up volumes, I think it is best practice to delete PV objects if users know volumes of these PVs are already cleaned or deleted if they want to clean these stale PV objects. And it's totally ok to not delete them if they are not sure. In some cases, volumes are not able to be deleted because they are not mounted. After mount points are fixed by users, PV objects will be able to be recycled by provisioner.

Conversely, I think we should check mount point existence on deletion because if no mount point exists on the directory, we actually didn't clean local volume and should not delete PV objects.

PR #126, opened 2019-07-05
auto remove the pv with hardware problems

yuyulei:
What this PR does?

When the pv has a hardware problem,
1. If the pv is Bound state, some pod is using it, after migrating the affected pod to another node, we can delete pvc, so pv will beReleased state.
2. If the pv is Available state, after recycling the hardware, we have to manually delete broken pv.

Now, provisioner can do case 1, but can not handle case 2 automatically.
My pr is to solve it. If OK, I will add unit test.

### DIFF
--- a/pkg/deleter/deleter.go
+++ b/pkg/deleter/deleter.go
@@ -69,9 +69,30 @@ func NewDeleter(config *common.RuntimeConfig, cleanupTracker *CleanupStatusTrack
 // delete them
 func (d *Deleter) DeletePVs() {
 	for _, pv := range d.Cache.ListPVs() {
-		if pv.Status.Phase != v1.VolumeReleased {
+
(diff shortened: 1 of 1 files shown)


yuyulei:
@cofyc @saad-ali

cofyc:
hi, for security, we cannot simply delete the PV object if we failed to clean the directory.
the hardware problem may be a temporary issue, [data in the volume may be leaked to other programs](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/6#issuecomment-484062771). Currently, admins must intervene when there are some hardware issues.

Possible solutions:

- fix the hardware issue and wait for the provisioner to clean the 
- delete the PV object manually and make sure the data is cleaned (e.g. remove the hardware)

PR #133, opened 2019-07-18
Host file name label

MOZGIII:
This PR adds a label to the PV with the host file name that the PV corresponds to.

Example:

```yaml
local-storage-host-file: "symlink1"
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -81,6 +81,9 @@ const (
 
 	// DefaultVolumeMode is the default volume mode of created PV object.
 	DefaultVolumeMode = "Filesystem"
+
+	// HostFileLabelKey is a label key for a label holding host file name of the PV.
+	HostFileLabelKey = "local-storage-host-file"
 )
 
 // UserConfig stores all the user-defined parameters to the provisioner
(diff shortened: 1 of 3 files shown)


MOZGIII:
I'm in the process of signing CLA (https://identity.linuxfoundation.org/projects/cncf seems to be broken currently). In the mean time we can discuss this PR.

MOZGIII:
I've signed CLA.

msau42:
Hi can you describe the use case you're trying to solve?  The file name is already accessible in the PV object at PV.Spec.Local.Path

MOZGIII:
I was looking for a way to specify to which PVs to bind at PVC template in StatefulSet. Though this would be the way...

cofyc:
it's better to classify your PVs by using storage classes. for example, if you have local volumes at two different places on your nodes (e.g. one is for HDD, one is for SSD), you can configure two local storage classes.

MOZGIII:
I agree, though is does not fit for my use case. I need PVs to be single purpose (i.e. for `my_database_1`) - while the storage class is supposed to be SSD / HDD / Hybrid etc... The goal is to partition the actual drives by need basis, and to have them properly mapped to PVCs. It would not be very handy to create a storage class per PV if I just need persistent names.

For example - I have a 120 GB SSD, and I make a 10 GB partition for a database 1 and 15 GB for database 2. I then need to crate PVC manually to avoid any conflicts, and it'd be great if I use proper names.

MOZGIII:
So, this went nowhere, what a shame

PR #135, opened 2019-08-10
controller/discoverer: add readiness check

JulienBalestra:
This PR allows to establish a healthcheck over a local volume provisioner pod.

This allows other controllers in the cluster to actually validate if a node has all it's system (daemonset) components up and running.

An immediate example of consumer would be the cluster-autoscaler to improve its capability with local volumes.

As a first step I added a readiness check for the `discoverer.DiscoverLocalVolumes()` call.
These checks could be easily be extended in other packages.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -919,7 +919,7 @@
   version = "kubernetes-1.14.0"
 
 [[projects]]
-  digest = "1:34093094960df1db1c921bea4dc054663f4bce54cf41de5c240f0864e23140c7"
+  digest = "1:1967352ea3ad2426f023b6366bc965ec6581039cbf54088865d678b23c3e8425"
   name = "k8s.io/apiserver"
   packages = [
     "pkg/apis/audit",
@@ -932,6 +932,7 @@
     "pkg/authentication/user",
     "pkg/authorization/authorizer",
     "pkg/features",
+    "pkg/server/healthz",
     "pkg/storage",
     "pkg/storage/etcd",
     "pkg/util/feature",
@@ -1575,6 +1576,7 @@
     "k8s.io/apimachinery/pkg/util/sets",
     "k8s.io/apimachinery/pkg/util/uuid",
     "k8s.io/apimachinery/pkg/util/wait",
+    "k8s.io/apiserver/pkg/server/healthz",
     "k8s.io/client-go/informers",
     "k8s.io/client-go/kubernetes",
     "k8s.io/client-go/kubernetes/fake",
--- a/docs/provisioner.md
+++ b/docs/provisioner.md
@@ -43,10 +43,14 @@ The basic components of the provisioner are as follows:
   The discovery and deleter run serially to simplify synchronization with the cache
   and create/delete operations.
 
-## Prometheus Metrics
+## Monitoring
 
-The metrics are exported through the Prometheus golang client on the HTTP
-endpoint `/metrics` on the listening port (default 8080).
+A dedicated HTTP server (default listening on 0.0.0.0:8080) exposes metrics and
+readiness state. 
+
+### Metrics
+
+The metrics are exported through the Prometheus golang client on the path `/metrics`.
 
 | Metric name                                                   | Metric type | Labels                                                                                                                                                                             |
 | ----------                                                    | ----------- | -----------                                                                                                                                                                        |
@@ -61,3 +65,11 @@ endpoint `/metrics` on the listening port (default 8080).
 | local_volume_provisioner_proctable_running                    | Gauge       |                                                                                                                                                                                    |
 | local_volume_provisioner_proctable_failed                     | Gauge       |                                                                                                                                                                                    |
 | local_volume_provisioner_proctable_succeeded                  | Gauge       |                                                                                                                                                                                    |
+
+### Readiness
+
+The readiness state is exposed on the path `/ready`.
+
+The state become ready when discovered local volumes are successfully created.
+
+Note that if there is no disk to create, the state will be marked as ready.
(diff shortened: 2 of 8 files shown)


Gopkg.lock
```diff
@@ -254,6 +254,14 @@
   revision = "20f1fb78b0740ba8c3cb143a61e86ba5c8669768"
   version = "v0.5.0"
 
+[[projects]]
+  branch = "master"
+  digest = "1:b669bf7ea147e4d75e097e89f45c1e7de52e79dc913bbbae94b79177f7b075b6"
+  name = "github.com/heptiolabs/healthcheck"
```
cofyc: hi, thanks for contributing this feature! because this is a k8s project, for consistent behavior, what about using this package instead? `https://godoc.org/k8s.io/apiserver/pkg/server/healthz`

cofyc:
I've opened a PR for build issue: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/137

pkg/discovery/discovery.go
```diff
@@ -184,9 +189,26 @@ func generateVolumeNodeAffinity(node *v1.Node) (*v1.VolumeNodeAffinity, error) {
...
+	var discoErrors []error
 	for class, config := range d.DiscoveryMap {
-		d.discoverVolumesAtPath(class, config)
+		err := d.discoverVolumesAtPath(class, config)
+		if err != nil {
+			discoErrors = append(discoErrors, err)
+		}
 	}
+	d.readySync.Lock()
+	d.ready = discoErrors == nil
```
msau42: Do we need to store an array of errors? Can it just be a bool?

pkg/discovery/discovery.go
```diff
@@ -209,31 +231,31 @@ func (d *Discoverer) getMountOptionsFromStorageClass(name string) ([]string, err
...
 
-func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConfig) {
+func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConfig) error {
 	klog.V(7).Infof("Discovering volumes at hostpath %q, mount path %q for storage class %q", config.HostDir, config.MountDir, class)
 
 	reclaimPolicy, err := d.getReclaimPolicyFromStorageClass(class)
 	if err != nil {
 		klog.Errorf("Failed to get ReclaimPolicy from storage class %q: %v", class, err)
-		return
+		return err
```
msau42: I prefer the error handling style where the callers log the errors (to avoid double logging at every layer).

cofyc:
yes, agree, the metrics section is [here](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/provisioner.md#prometheus-metrics)

pkg/discovery/discovery.go
JulienBalestra: Yes in this case it could be 👍

pkg/discovery/discovery.go
JulienBalestra: Would you like me to move all the different logging lines in this function to the caller:
```go
(d *Discoverer) DiscoverLocalVolumes()
```
Don't you think we're going to loose some details about the error returned if we just log in the caller ?
It looks like all logs messages add more context to the returned error. Should we wrap them with this additional context ?

msau42: How about `return fmt.Error("Failed to get ReclaimPolicy" ....)`

JulienBalestra: I've done the changes, let me know if it matches what you proposed 👍

docs/provisioner.md
```diff
@@ -61,3 +65,7 @@ endpoint `/metrics` on the listening port (default 8080).
 | local_volume_provisioner_proctable_running                    | Gauge       |                                                                                                                                                                                    |
 | local_volume_provisioner_proctable_failed                     | Gauge       |                                                                                                                                                                                    |
 | local_volume_provisioner_proctable_succeeded                  | Gauge       |                                                                                                                                                                                    |
+
+### Readiness
+
+The readiness state is exposed on the path `/ready`.
```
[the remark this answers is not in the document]
JulienBalestra: @msau42 I've done some adjustment there. Could you tell me what do you think ?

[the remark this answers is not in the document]
msau42: I don't see any update, did you push the change?

docs/provisioner.md
```diff
@@ -61,3 +65,9 @@ endpoint `/metrics` on the listening port (default 8080).
...
+
+The state become ready when discovered local volumes are successfully created.
```
msau42: one more nit clarification: on the initial pass

msau42: Would it be worth calling out that if there were no disks found on the initial pass, then it will also be ready?

JulienBalestra: @msau42 it's not on the initial pass only. For example if the RBAC is removed the pv creation will fail so the pod will become notReady.

This is what I had in mind, would you like me to change this to become ready on the initial pass only ?

msau42: Hm ok I think the rbac removal afterwards is a valid scenario of it becoming not ready.  The main scenario I was thinking of was:

1. Start provisioner before any disks were added/mounted
2. Provisioner sets ready = true because there were no disks to discover
3. After disks were added/mounted, now there are issues, and readiness gets set to false.

Is it ok for the provisioner to have been ready in step 2?  I think it's fine (and I don't see an easy way to really detect this scenario), but do you think it's worth clarifying that it might be ready even if there were no disks and/or no work to do (ie all the disks were already discovered)?

JulienBalestra: @msau42 Agreed, I'm going to add some information about this in the docs 👍

msau42:
/lgtm
Thanks!

PR #138, opened 2019-08-15
fix docker build arguments

### LINKED ISSUE #95: local-volume-provisioner for arm

cofyc:
fix #95 

> In a Dockerfile, FROM starts a new build-stage, and any property from the previous stage (if any) is reset.

- https://github.com/docker/distribution/issues/2459#issuecomment-348848758
- https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

### DIFF
--- a/deployment/docker/Dockerfile
+++ b/deployment/docker/Dockerfile
@@ -12,17 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG OS=linux
 ARG ARCH=amd64
 ARG GOVERSION=1.11.1
 FROM golang:$GOVERSION-alpine as builder
+ARG OS=linux
+ARG ARCH=amd64
 WORKDIR /go/src/sigs.k8s.io/sig-storage-local-static-provisioner
 ADD . .
+RUN echo "OS: $OS ARCH: $ARCH"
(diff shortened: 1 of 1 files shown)


deployment/docker/Dockerfile
```diff
@@ -12,17 +12,21 @@
...
-ARG OS=linux
-ARG ARCH=amd64
 ARG GOVERSION=1.11.1
+ARG ARCH=amd64
 FROM golang:$GOVERSION-alpine as builder
+ARG OS=linux
+ARG ARCH=amd64
 WORKDIR /go/src/sigs.k8s.io/sig-storage-local-static-provisioner
 ADD . .
+RUN echo "OS: $OS ARCH: $ARCH"
```
cofyc: docker build does not show replaced variables. I add this to show current values.

deployment/docker/Dockerfile
```diff
@@ -12,17 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG OS=linux
 ARG ARCH=amd64
```
msau42: Do we still need these 2 args?

cofyc: `ARG OS=linux` is not needed, so deleted. `ARG ARCH=amd64` is required because it's used in `FROM k8s.gcr.io/debian-base-${ARCH}:0.4.1` instruction.

PR #142, opened 2019-08-29
disable non-amd64 arches

cofyc:


hack/e2e.go
```diff
@@ -92,6 +92,8 @@ func wait(cmd string, args ...string) error {
 	c := exec.Command(cmd, args...)
 	c.Stdout = os.Stdout
 	c.Stderr = os.Stderr
+	c.Env = append(c.Env, os.Environ()...)
+	c.Env = append(c.Env, "GO111MODULE=on")
```
cofyc: used to resolve kubetest compile issue

PR #148, opened 2019-09-11
add prefix check for adding pv to cache

yuyulei:
If node recycled for some reasons(e.g. add cpu/mem/ip capacity, hardware error), then the node rejoined the cluster, node name not changes, but uid changes.

The problem is that some pvs is created before node recycling, now they not match node uid.

```
pv.kubernetes.io/provisioned-by: local-volume-provisioner-<node name>-<old uid>
```

So provisiner will not add these pvs to cache. 

Now I want use `UserConfig.UseNodeNameOnly` to ignore node uid info when create pv. Further, I also hope the provisioner can add old pvs to cache. At last, I make the pr.

### DIFF
--- a/pkg/populator/populator.go
+++ b/pkg/populator/populator.go
@@ -17,11 +17,12 @@ limitations under the License.
 package populator
 
 import (
-	"k8s.io/klog"
-	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
+	"strings"
 
 	"k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
 )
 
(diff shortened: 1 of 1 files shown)


yuyulei:
@cofyc

pkg/populator/populator.go
```diff
@@ -72,7 +73,7 @@ func (p *Populator) handlePVUpdate(pv *v1.PersistentVolume) {
 			if !found {
 				return
 			}
-			if provisioner == p.Name {
+			if provisioner == p.Name || strings.HasPrefix(provisioner, p.Name) {
```
yuyulei: may I  use "strings.HasPrefix(provisioner, p.Name) " to replace "provisioner == p.Name "?

yuyulei:
@cofyc

cofyc:
after changing `UseNodeNameOnly` to `true` from `false`, how about writing a script to migrate existing PVs to the new provisioner

PR #149, opened 2019-09-26
support to run e2e with kind

### LINKED ISSUE #113: Start Kubernetes via minikube
minikube is a simpler way to start a local kubernetes than `hack/local-up-cluster.sh`.

cofyc:
fixes #113 

this makes it easy to run e2e tests in a multi-node cluster locally

depends on https://github.com/kubernetes/test-infra/pull/14477
depends on https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/155 #157

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -40,12 +40,12 @@ Usage: hack/e2e.sh [-h] -- [extra kubetest args]
 Environments:
 
     ARTIFACTS                           directory where job artifacts can be dumped
-    PROVIDER                            local/gce/gke (detect automatically if not specified)
+    PROVIDER                            local/gce/gke/skeleton (detect automatically if not specified)
     GCP_ZONE                            (for gce/gke) GCP zone
     GCP_PROJECT                         (for gce/gke) GCP project
     EXTRACT_STRATEGY                    kubetest extract strategy, see k8s.io/test-infra/kubetest/README.md for explanation
     KUBERNETES_SRC                      if specified, kubetest will skip extracting kubernetes src from GCS, use it instead
-    DEPLOYMENT                          none/bash/gke
+    DEPLOYMENT                          none/bash/gke/local/kind
     GKE_ENVIRONMENT                     (gke only) test/staging/prod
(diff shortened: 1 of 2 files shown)


cofyc:
xref: https://github.com/kubernetes/test-infra/pull/14094#issuecomment-535922595

msau42:
Do we deploy with hostpath to /dev?  @BenTheElder mentioned the other day that devices in kind will passthrough to the host, which gets shared by multiple prow jobs.  Unsure if our block tests may cause devices to leak on the hosts?

cofyc:
no hostpath for `/dev`. loop devices are created and symlinked into discovery directory for testing.

cofyc:
ah, I see the problem is `/dev` is shared by the host and kind nodes containers. There is no namespace support for loop devices, so devices created in kind nodes are shared by all containers.

It's not suitable to run devices tests with kind. Current gke/gce prow jobs are enough and no need to run e2e tests with kind in the prow. We can add this as an alternative way to run e2e tests locally.

hack/e2e.sh
```diff
@@ -68,7 +68,12 @@ Examples:
 
   Note that current kubetest needs root permission to cleanup.
 
-3) To run against cluster with GCE provider locally
+3) TO run against a cluster started by kind
```
msau42: Can you add a note about the limitation with block devices here?

BenTheElder:
> Do we deploy with hostpath to /dev? @BenTheElder mentioned the other day that devices in kind will passthrough to the host, which gets shared by multiple prow jobs. Unsure if our block tests may cause devices to leak on the hosts?

Yeah I haven't had a chance to dig into this further but AIUI privileged containers will pass through `/dev` and then this will get passed down by repeatedly... with no namespacing.

I'm not sure how to handle this better other than trying not to run these.
It would be helpful if there was an easy regex to skip any test using /dev (is there?)

hack/e2e.sh
cofyc: done

msau42:
> It would be helpful if there was an easy regex to skip any test using /dev (is there?)

We can probably skip on "block" or something like that.

PR #155, opened 2019-11-07
upgrade to latest debian-base version

cofyc:


test/e2e/e2e.go
```diff
@@ -126,12 +127,14 @@ var _ = ginkgo.SynchronizedAfterSuite(func() {
...
 	// Register framework flags, then handle flags and Viper config.
-	framework.HandleFlags()
+	framework.RegisterCommonFlags()
+	framework.RegisterClusterFlags()
 	framework.AfterReadingAllFlags(&framework.TestContext)
 }
 
 // RunE2ETests runs e2e tests.
 func RunE2ETests(t *testing.T) {
+	flag.Parse()
```
cofyc: prow base image is upgraded to use go1.13.4, there is a breaking change in go1.13 testing package (xref: https://golang.org/doc/go1.13#testing) that testing flags are not registered in the initialization phase, so we can't prase the flags in init() now.

test/e2e/e2e_test.go
```diff
@@ -992,6 +993,17 @@ func deletePodAndPVCs(config *localTestConfig, pod *v1.Pod) error {
...
+func init() {
+	// Register framework flags, then handle flags and Viper config.
+	framework.RegisterCommonFlags()
+	framework.RegisterClusterFlags()
+	framework.AfterReadingAllFlags(&framework.TestContext)
+}
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+}
```
cofyc: prow base image is upgraded to use go1.13.4, there is a breaking change in go1.13 testing package (xref: https://golang.org/doc/go1.13#testing) that testing flags are not registered in the initialization phase, so we can't prase the flags in init() now.

PR #160, opened 2019-12-06
metrics: report the capacity in bytes of the local volumes discovered

JulienBalestra:
This PR allows to expose a metrics of the cumulated size, in bytes, of the local volumes discovered.

I had some hesitation by setting this with additional labels in a `GaugeVec` like the `pvName` but I thought it would be more complex to reconcile any missing/gone volume.


pkg/metrics/metrics.go
```diff
@@ -47,6 +47,14 @@ var (
 		},
 		[]string{"mode"},
 	)
+	// PersistentVolumeSizeBytes is used to collect the size of the local volumes discovered.
+	PersistentVolumeSizeBytes = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Subsystem: LocalVolumeProvisionerSubsystem,
+			Name:      "persistentvolume_size_bytes",
```
cofyc: - I guess `s/size/capacity/' the users can know it's the size of capacity
- better break down by mode, see [PersistentVolumeDiscoveryTotal](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/6632c7dc895f11318747663a519aa400e6d96fc7/pkg/metrics/metrics.go#L42-L49) as an example, then end-users can know the total capacity in size for all volumes types (by summing) and also for each mode

pkg/metrics/metrics.go
```diff
@@ -47,6 +47,14 @@ var (
...
+			Help:      "Persistent volumes discovered in bytes.",
```
cofyc: describe it's the capacity of all discover Persistent volumes in help

pkg/metrics/metrics.go
JulienBalestra: I've done the requests changes 👍

cofyc: LGTM, one last nit: update metrics docs [here](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/provisioner.md#metrics)

JulienBalestra:
@cofyc I've done the needed changes and re-ordered the metrics alphabetically.
Let me know if everything is fine 👌

cofyc:
/lgtm
Thanks!

PR #169, opened 2020-02-05
docs on configuration and upgrading

cofyc:
docs on configuration and upgrading

Signed-off-by: Yecheng Fu <fuyecheng@pingcap.com>


cofyc:
fixed a typo

PR #130, opened 2019-07-16
Add notes about using Filesystem on block devices

mattymo:
Fixes #129

PR #170, opened 2020-02-05
add notes about volumeMode and fix TOC

cofyc:
follow-up of #130 
Signed-off-by: Yecheng Fu <fuyecheng@pingcap.com>


msau42:
/lgtm
Thanks for cleaning up the docs!

PR #174, opened 2020-02-18
Remove old cleanup status before create new pv

yuyulei:
**Ⅰ. Describe what this PR does**

fix a bug, the bug will make that pv created has dirty data.

**II. How to reproduce the BUG**

At first, privisioner  will loop check pv every 10 seconds, then I describe the steps by timeline.

00:00 provisioner find a pvX with "Released" status, put the pvX cleanup status to Table, and mark it "Running"

00:01 the async task completes cleanup work quickly, and mark it "Succeed"

**00:02~00:09 pvX is deleted by some people or itself, and provisioner cache removes the pv. BUT the cleanup status of pvX stays in the Table(ProcTableImpl)**

00:10 provisioner recreate the pvX

some hours laster, the status of pvX becomes "Released"

10:00 provisioner find pvX "Released", and try to get cleanup status of pvX from Table. Unfortunately, provisioner find the "Succeed" from "00:02~00:09", then just delete pv without cleanuping  So the data leaks. 

Other cases also can recreate it. For example, when apiserver has problems, provisioner delete a pv successfully, but provisioner watch the delete event after some seconds, maybe more than 10 seconds. 

**III. How fix** 

I just clear the cleanup status when the pv recreated.

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -307,6 +307,13 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 			continue
 		}
 
+		// remove old cleanup status
+		_, _, err = d.CleanupTracker.RemoveStatus(pvName, usejob)
+		if err != nil {
+			klog.Errorf("expected status exists and fail to remove cleanup status for pv %s", pvName)
+			continue
+		}
+
 		mountOptions, err := d.getMountOptionsFromStorageClass(class)
 		if err != nil {
 			discoErrors = append(discoErrors, fmt.Errorf("failed to get mount options from storage class %s: %v", class, err))


cofyc:
it's a serious bug, I think we should have some tests. do you know how to reproduce this issue?

yuyulei:
@cofyc follow "II. How to reproduce the BUG" mentioned from my description.

cofyc:
I mean the steps to reproduce the same issue again. Do the normal operations will trigger the issue or some conditions must be met? e.g. PV must be deleted within the specified time.

yuyulei:
00:00 first find the released pv
00:01~00:09 delete pv manually(this is the special time, you must meet: 1. pv cleanup status has become SUCCEED, 2: before next loop)
00:10 provisioner  create the pv with same name 

01:00 first pod bind to the pv, that's OK
01:10 first pod and its pvc are deleted
provisioner will recreate the pv 

02:00 second pod bind to the pv again, that's not OK, you can find the dirty data.

cofyc:
Thanks for your explanation!

I see the root cause is that if the PV is deleted and recreated again in one controller loop, the status of the delete job will not be cleared in `Deleter.deletePV` because the PV is not Released in the next loops. Then when the PV is populated with some data and released again, the provisioner will skip the deletion because a stale `Succeed` status will be found.

https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/85713966699994d7de71eeffcb7fcbfb9673621b/pkg/deleter/deleter.go#L156-L160

cofyc:
/lgtm
/approve

Thanks for the fix! It looks good to me to clear the status of the delete job when PV is recreated.

PR #179, opened 2020-02-27
Helm chart rework v2

HaveFun83:
Did some refactoring on the helm chart to match the helm best practices

* split deployment files into "single objects"
* join values from rbac and serviceAccount
* removed namespace creation. As helm v2 will create one during install and on helm v3 the  namespace must already created before install.
* removed common.configMapName . This will be generated from the release-name
* removed daemonset.name. This will be generated from the helm release-name
* removed daemonset.serviceAccount This will be generated from the helm release-name
* changed metadata.labels
* rework the Prometheus servicemonitor creation
* added generation / verify for helm v2 and helm v3

Please have a look. Feedback is welcome
Thanks

@cofyc sorry i opened a new pr as i messed up the old one


HaveFun83:
@cofyc any news here?

helm/examples/baremetal-namespace.yaml
```diff
@@ -1,10 +0,0 @@
-common:
```
cofyc: why this is removed?

helm/provisioner/values.yaml
```diff
@@ -3,17 +3,21 @@
 #
```
cofyc: the format of the new values LGTM, but we should make the new chart backward compatible with old values yaml as possible as we can, e.g.

- `daemonset.name` override the name of DaemonSet if present
- `common.configMapName` overrides the name of ConfigMap if present
- `common. podSecurityPolicy` overrides the `rbac.pspEnabled` if present
- etc.

then, people can use the old values file with the new chart and don't worry about any problems during upgrading.

hack/lib.sh
```diff
@@ -36,6 +38,11 @@ function hack::verify_helm() {
     if test -x "$HELM_BIN"; then
         local v=$($HELM_BIN version --short --client | grep -o -P '\d+.\d+.\d+')
         [[ "$v" == "$HELM_VERSION" ]]
+        return
+		fi
+		if test -x "$HELM2_BIN"; then
+        local v=$($HELM2_BIN version --short --client | grep -o -P '\d+.\d+.\d+')
+        [[ "$v" == "$HELM2_VERSION" ]]
```
cofyc: it's not right to check helm2 and helm3 in one function. if `$HELM_BIN` exist, `$HELM2_BIN` will not be checed.

you can add another two functions for `helm2`, e.g. hack::verify_helm2, hack::install_helm2.

hack/verify-generated.sh
```diff
@@ -35,13 +35,32 @@ if [ $ret -ne 0 ]; then
...
+}
+
+# check examples helm3
 function test_values_file() {
     local input="examples/$1"
     local expected="generated_examples/$1"
     local tmpfile=$(mktemp)
     trap "test -f $tmpfile && rm $tmpfile || true" EXIT
-    $HELM_BIN template ./provisioner -f examples/$f > $tmpfile
+    #echo "$HELM_BIN"
```
cofyc: nit: unnecessary comments

helm/examples/baremetal-namespace.yaml
HaveFun83: because:
>removed namespace creation. As helm v2 will create one during install and on helm v3 the namespace must already created before install.

hack/verify-generated.sh
HaveFun83: removed

helm/provisioner/values.yaml
HaveFun83: I can do this but it means duplication in the value / chart template files which is not a good approach in my eyes. 
Please have look in the helm/Readme.md i added a section about upgrading.
During tests within vagrant i saw no issues during upgrading from old to new chart version.

hack/lib.sh
HaveFun83: Done with new commit

helm/provisioner/values.yaml
cofyc: ok

helm/examples/baremetal-namespace.yaml
cofyc: got it

cofyc:
/lgtm
/approve

Great work!

HaveFun83:
@cofyc  big thanks for your support and review.

PR #187, opened 2020-05-20
feat: add namePattern field in discovery PV

andyzhangx:
This PR is to add a `namePattern ` field, with following config, only `/dev/nvm*` device names would be discovered, by default `namePattern ` field is empty, it won't do any matching, so it's backward compatible.

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: local-provisioner-config
  namespace: default
data:
  storageClassMap: |
    fast-disks:
       hostDir: /dev
       mountDir:  /dev
       blockCleanerCommand:
         - "/scripts/shred.sh"
         - "2"
       volumeMode: Filesystem
       fsType: ext4
       namePattern: "nvme*"
```

/assign @cofyc

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -278,6 +278,16 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 	var discoErrors []error
 	var totalCapacityBlockBytes, totalCapacityFSBytes int64
 	for _, file := range files {
+		if config.NamePattern != "" {
+			matched, err := filepath.Match(config.NamePattern, file)
+			if err != nil {
+				return err
+			}
+			if !matched {
+				klog.Infof("file(%s) under(%s) does not match pattern(%s)", file, config.MountDir, config.NamePattern)
+				continue
+			}
+		}
 		startTime := time.Now()
 		filePath := filepath.Join(config.MountDir, file)
 		volMode, err := common.GetVolumeMode(d.VolUtil, filePath)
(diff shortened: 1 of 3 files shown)


pkg/discovery/discovery.go
```diff
@@ -278,6 +279,10 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 	var discoErrors []error
 	var totalCapacityBlockBytes, totalCapacityFSBytes int64
 	for _, file := range files {
+		if config.DeviceNameMatching != "" && !strings.Contains(file, config.DeviceNameMatching) {
+			klog.Infof("file(%s) under(%s) does not match %s", file, config.MountDir, config.DeviceNameMatching)
+			continue
+		}
```
cofyc, line 285: how about making it generic?

1) rename `DeviceNameMatching` to `NamePattern` (filesystem mount points are also matched here)
2) use [filepath.Match](https://golang.org/pkg/path/filepath/#Match) to match on file name

cofyc: in your case, the pattern will be `nvm*`, the same pattern used in the shell.

andyzhangx: sounds good, will change the PR soon

andyzhangx: thanks, it's done. To make sure it's backward compatible, when `namePattern` is empty string, it would be invoke `filepath.Match`, so empty == "*" here.

cofyc: LGTM, btw can you add a helm value for this configuration (see [example](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/tree/master/helm#configurations), e.g. `classes.[n].fsType`)? you can do it in a separate PR.

andyzhangx: sure, thanks for the review

cofyc:
/lgtm
/approve
Thanks!

PR #189, opened 2020-05-22
feat: add helm support on namePattern field

andyzhangx:
a follow up with PR: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/187

/assign @cofyc


cofyc:
/lgtm
/approve
Thanks again!

PR #200, opened 2020-06-17
Fix blkdiscard script

rbtcollins:
blkdiscard with -z is precisely equivalent to the dd_zero script: rather
than discarding sectors, they are written to, causing more writes to the
drive, and taking a long time.

Quoting the man page:
"""
       -z, --zeroout
              Zero-fill rather than discard.
"""

Signed-off-by: Robert Collins <robert.collins@cognite.com>

**What type of PR is this?**

/kind bug

This PR makes the blkdiscard script actually discard blocks on SSD/NVMe drives, rather than writing \00's.

Fixes #

**Release note**:
```

```

/sig storage-local-static-provisioner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ helm chart refactoring:
 - **Action required**\
   As the helm-chart structure changed the already running pod will be recreated during upgrade. Documentation can be found under [helm/README.md](./helm/README.md). Compare your existing values with the new chart parameter before upgrade.  [#171](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/171)
 
+- blkdiscard.sh no longer zeros disks.
+  This script was passing the -z option to blkdiscard which meant it was not
+  performing discards. This has been fixed. If you desire zeroing, rather than
+  block discarding, please switch to dd_zero.sh.
+
 # [v2.3.4](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/releases/tag/v2.3.4)
 
 Image updates:
--- a/deployment/docker/scripts/blkdiscard.sh
+++ b/deployment/docker/scripts/blkdiscard.sh
@@ -31,4 +31,4 @@ fi
 validateBlockDevice
 
 echo "Calling blkdiscard"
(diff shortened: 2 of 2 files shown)


cofyc:
hi, thanks!
zeroing a large disk takes a long time, but I'm afraid someone might rely on this secure but slow script. 
can you add a new script, e.g. `blkdiscard_quick.sh`?

rbtcollins:
I can; I suggest that we delete the old script if we are going to do that, because it isn't actually discarding - and the point of discarding is to maintain SSD performance - so it is really a foot-gun today. Sound ok?

cofyc:
can you add a change log in `CHANGELOG.md` (under HEAD section)?  

people who want zeroing out the disk can use other scripts, e.g. `dd_zero.sh`

cofyc:
> I can; I suggest that we delete the old script if we are going to do that, because it isn't actually discarding - and the point of discarding is to maintain SSD performance - so it is really a foot-gun today. Sound ok?

hi, I mean it's ok to update `blkdiscard.sh` as you suggested, but we need to add a changelog to tell people if they use `blkdiscard.sh` to zero out disks, they can use `dd_zero.sh` instead.

rbtcollins:
Oh right, how does this look

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ helm chart refactoring:
 - **Action required**\
   As the helm-chart structure changed the already running pod will be recreated during upgrade. Documentation can be found under [helm/README.md](./helm/README.md). Compare your existing values with the new chart parameter before upgrade.  [#171](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/171)
 
+- blkdiscard.sh no longer zeros disks.
+  This script was passing the -z option to blkdiscard which meant it was not
+  performing discards. This has been fixed. If you desire zeroing, rather than
+  block discarding, please switch to switch to dd_zero.sh.
```
cofyc, line 10: nit: `s/switch to switch to/switch to/`

cofyc: we can merge once this typo is fixed :)

cofyc:
/lgtm
/approve
Thanks!

Issue #107, opened 2019-05-31
Initialization failed for volume


spencergilbert:
Error: `MountVolume.NewMounter initialization failed for volume "local-pv-7f70b248" : path "/mnt/local/vol1" does not exist`

I could certainly be doing something pretty silly here but I'm unable to mount volumes anymore. I had tested this earlier in the year and didn't have issues so I'm not sure what I'm missing. I've attached the manifest I used.

[templated_manifest.txt](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/files/3239650/templated_manifest.txt)

cofyc:
The kubelet emitted this error log because the path `/mnt/local/vol1` didn't exist on the node. I guess `/mnt/local/vol1` is the path of local PV used by the pod. You need to check the path on that node.

spencergilbert:
Off the host:
`root@halfling-fidel:~# cd /mnt/local/vol1/
root@halfling-fidel:/mnt/local/vol1# pwd
/mnt/local/vol1`

cofyc:
Is your kubelet running in a container?

spencergilbert:
Ah, yes. We're using Rancher so it is containerized.

cofyc:
If the kubelet is running in a container, it cannot access the host filesystem at the same path. You must adjust `hostDir` to the correct path in the kubelet container. 

For example, if the host filesystem is mounted at `/rootfs` in the kubelet container, you need to set `hostDir` to `/rootfs/mnt/local`.

cofyc:
By the way, you can also add a bind for local volumes in kubelet container. e.g.

```
        "HostConfig": {
            "Binds": [
                "/mnt/local:/mnt/local"
            ],
```

spencergilbert:
I really should have documented all this stuff I did when it worked the first time 🙃 

Thanks @cofyc

meyerbro:
Hey @spencergilbert, how did you solve this? I'm also using rancher and I didn't find a hostDir other than "/" on my case...

meyerbro:
@cofyc, any guess on this one? Thanks!

spencergilbert:
Just as above since the kubelet is dockerized you need to pass additional bind config into the cluster yaml config, i've been matching ie. create pv at /mnt/local on worker node bind config as /mnt/local:/mnt/local

cofyc:
[...]

PR #215, opened 2020-08-19
add notes for rancher extra binds

cofyc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
/kind documentation
> /kind test
> /kind failing-test
> /kind feature
> /kind flake

https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/107#issuecomment-521478232

I was asked several times about how to use local volume provisioner in Rancher. I think we should explicitly document it.

Fixes #

**Release note**:
```

```

### DIFF
--- a/docs/faqs.md
+++ b/docs/faqs.md
@@ -195,3 +195,7 @@ them to be discovered again.
 Another solution is to add a bind in the kubelet deployment configuration to
 mount the parent directory of local volumes into the kubelet container at the
 same path. This requires restarting the kubelet container.
+
+For [Rancher](https://rancher.com) clusters, users must configure additional
+local volumes (or the parent directory) via [Extra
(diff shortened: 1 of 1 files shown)


docs/faqs.md
```diff
@@ -195,3 +195,8 @@ them to be discovered again.
 Another solution is to add a bind in the kubelet deployment configuration to
 mount the parent directory of local volumes into the kubelet container at the
 same path. This requires restarting the kubelet container.
+
+A famous Kubernetes distribution that runs the kubelet in a container is
```
msau42, line 199: simplify the first part of the sentance to something like:

"For Rancher clusters, users must configure ..."

cofyc: updated

PR #219, opened 2020-09-23
improv: Add opt-out for `/dev` volume in the chart

maybe-sybr:
*What type of PR is this?**
/kind feature

For filesystem only based provisioning, `/dev/` shouldn't need to be
mounted so we add a new value to opt-out of defining it in the chart.

**Release note**:
```

```


maybe-sybr:
The CLA should be fine now - it went green in my other PR

cofyc:
/ok-to-test
Thanks!
btw, can you update the docs here? https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/tree/master/helm#configurations

maybe-sybr:
Should be good now, @cofyc - thanks for the quick review :)

PR #235, opened 2021-02-06
change log level to V5 when dicovering path not match pattern

### LINKED ISSUE #216: Reducing local-volume-provisioner log verbosity

I'm using container image `mcr.microsoft.com/k8s/local-volume-provisioner:v2.3.5-alpha1` and it is producing a lot of info logs that should probably be debug/trace. See attached logs in just a few minutes.

I'd like to be able to lower the verbosity of this container.

[container.log](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/files/5121789/container.log)

Be able to tune the log level using an arg/env variable. 

Also the attached log seems like a good candidate for debug/trace levels, not info.

I tried setting the `-v` arg with various levels, but still got the same output. 

**Additional context**

Excessive logs translates to more costs storing these logs, so would like to be able to reduce the logs as needed.

@andyzhangx

hellogdc:
**What type of PR is this?**

/kind feature

By now, Discoverer uses klog.Infof to print log when dicovering path not match pattern. This may lead to large logs, just as issuse #216 says. So i think it may be better to change it to klog.V(5).Infof. 

Fixes #216 

**Release note**:
```
None
```

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -284,7 +284,7 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 				return err
 			}
 			if !matched {
-				klog.Infof("file(%s) under(%s) does not match pattern(%s)", file, config.MountDir, config.NamePattern)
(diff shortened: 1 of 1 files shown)


cofyc:
/lgtm
/approve
Thanks!

PR #246, opened 2021-03-31
Duplicate volume guard

davidmccormick:
**What type of PR is this?**
/kind bug

The local-static-provisioner can create multiple separate kubernetes volumes pointing at the same underlying local host path through misconfiguration, e.g. if you accidentally configure different storageClasses with the same path in your configmap:

```
apiVersion: v1
kind: ConfigMap
data:
  filterPVsByNodeName: "yes"
  storageClassMap: |
    local-storage:
      hostDir: /k8svolumes
      mountDir: /k8svolumes
      blockCleanerCommand:
        - /scripts/fsclean.sh
      volumeMode: Filesystem
    local-storage-again:
      hostDir: /k8svolumes
      mountDir: /k8svolumes
      blockCleanerCommand:
        - /scripts/fsclean.sh
      volumeMode: Filesystem
```

This will result in the two volumes being created for each local path - one for each two storageClasses configured - both available and mountable by different pods. 😱

You will also get duplicates if you change the name of a configured storageClass associated with a given discovery path **after** the PVs have already been created with the old class.  Again, this could be caused by an error, or as part of a conscious attempt to reorganise storage (as was in our case).

Having duplicates of essential the **same** volume mounted in different customer pods is rarely desirable, opens up security concerns and will likely result in data loss for the pods colliding on the underlying filesystem.  A more cautious approach can be adopted without a high implementation cost.

This change aims to prevent a new volume from being created when any other **local-static-provisioner** volume already exists (of any storageClass) using the same local host path.  This will provide a sensible safe-guard against human configuration errors.  It also opens up the possibility of safer storage migrations between classes because the configuration changes can be rolled out without fear of disrupting any active volumes.  Volumes could then be migrated on a volume-by-volume basis and the new volumes will not be provisioned until the old volumes are deleted. 

**Implementation**

The provisioner already makes use of an informer cache of PVs which it uses to ensure that it does not repeatedly try and provision the same volume.  This change adds a method to the cache to return a list of names of PVs that are present on a given path and a check during the discovery will block the creation of a volume with an error message if it finds that any other volumes (of any storage class but that belong to the local provisioner) exist with the path.

The informer cache is synced before the provisioner runs discovery so we wont have a race upon start up between cache population and discovery.

If the provisioner is switched to `useNodeNameOnly: true` mode then we want it to match any pre-existing volumes that were configured with the nodename-uid annotation so that we can properly load these volumes into the cache and prevent duplicates.

NONE

**Release note**:

```
Users are prevented from creating new physical volumes for local static paths that are used in existing volumes.  An attempt to do so will return an error indicating the path and the existing volume which are preventing the new volume from being created.
```

### DIFF
--- a/pkg/populator/populator.go
+++ b/pkg/populator/populator.go
@@ -17,6 +17,8 @@ limitations under the License.
 package populator
 
 import (
+	"strings"
+
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
 
@@ -83,6 +85,11 @@ func (p *Populator) handlePVUpdate(pv *v1.PersistentVolume) {
 		if pv.Annotations != nil {
 			provisioner, found := pv.Annotations[common.AnnProvisionedBy]
 			if !found {
+				return
+			}
+			if p.UseNodeNameOnly && strings.HasPrefix(provisioner, p.Name) {
+				// The PV was created on this node
+				p.Cache.AddPV(pv)
 				return
 			}
 			if provisioner == p.Name {
--- a/pkg/populator/populator.go
+++ b/pkg/populator/populator.go
@@ -87,14 +87,24 @@ func (p *Populator) handlePVUpdate(pv *v1.PersistentVolume) {
 			if !found {
 				return
 			}
-			if p.UseNodeNameOnly && strings.HasPrefix(provisioner, p.Name) {
-				// The PV was created on this node
-				p.Cache.AddPV(pv)
-				return
-			}
 			if provisioner == p.Name {
 				// This PV was created by this provisioner
 				p.Cache.AddPV(pv)
+				return
+			}
+			if p.UseNodeNameOnly {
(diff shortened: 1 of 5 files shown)


pkg/populator/populator.go
```diff
@@ -85,6 +87,11 @@ func (p *Populator) handlePVUpdate(pv *v1.PersistentVolume) {
 			if !found {
 				return
 			}
+			if p.UseNodeNameOnly && strings.HasPrefix(provisioner, p.Name) {
+				// The PV was created on this node
+				p.Cache.AddPV(pv)
+				return
+			}
```
cofyc, line 94: I'm afraid we cann't rely on this, as one node name can be the prefix of another node name.

cofyc: This works only if `config.NodeLabelsForPV` is enabled, otherwise `kubernetes.io/hostname` label will not be added to PV object. However, it's a safety check.

PR #256, opened 2021-07-06
WIP: Adding callbacks to be executed before deleting released PVs

### LINKED ISSUE #255: Adding callbacks to be executed before deleting released PVs

Local Static Provisioner doesn't have a callback feature and doesn't clean up the symlinks.

We can make the [DeletePVs()](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/pkg/deleter/deleter.go#L70) and [deletePV()](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/3ea867f9f7151eda164c98d82da7f83e659e58e9/pkg/deleter/deleter.go#L131) a variadic function which can take additional functions that will get executed after devices are cleaned and before PVs are deleted in `deletePV()` function. 
`func (d *Deleter) DeletePVs(callbackFn ...func(pv *v1.PersistentVolume) error) { }`

When used as a library other projects can get the callback functions from Local Static Provisioner and pass it to `DeletePVs()`

example for symlink cleanup:
`func (d *Deleter) getCleanPVSymlinksFunc() func(pv *v1.PersistentVolume) error { }`

Creating a copy of `DeletePVs()` in [Local Storage Operator](https://github.com/openshift/local-storage-operator) which will also clean the symlinks 

**Additional context**

This is required by LSO for an easier cleanup of PVs and symlinks when LocalVolume and LocalVolumeSet objects are deleted.
https://github.com/openshift/enhancements/blob/master/enhancements/storage/cleanup-lso-symlink.md

rohan47:
**What type of PR is this?**

feature

DeletePVs can take functions that will be executed after the devices are
cleaned and before the PVs are deleted.
Added function for symlink cleanup.

Fixes #255 

**Release note**:
```

```

### DIFF
--- a/pkg/deleter/deleter.go
+++ b/pkg/deleter/deleter.go
@@ -67,7 +67,7 @@ func NewDeleter(config *common.RuntimeConfig, cleanupTracker *CleanupStatusTrack
 
 // DeletePVs will scan through all the existing PVs that are released, and cleanup and
 // delete them
-func (d *Deleter) DeletePVs() {
+func (d *Deleter) DeletePVs(deleteFn func(pv *v1.PersistentVolume) error) {
 	for _, pv := range d.Cache.ListPVs() {
 		if pv.Status.Phase != v1.VolumeReleased {
 			continue
@@ -82,7 +82,7 @@ func (d *Deleter) DeletePVs() {
 		case v1.PersistentVolumeReclaimDelete:
 			klog.V(4).Infof("reclaimVolume[%s]: policy is Delete", name)
 			// Cleanup volume
-			err := d.deletePV(pv)
+			err := d.deletePV(pv, deleteFn)
 			if err != nil {
 				mode, modeErr := d.getVolMode(pv)
 				if modeErr != nil {
@@ -128,7 +128,7 @@ func (d *Deleter) shouldRunJob(mode v1.PersistentVolumeMode) bool {
 	return mode == v1.PersistentVolumeBlock && d.RuntimeConfig.UseJobForCleaning
 }
 
-func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
+func (d *Deleter) deletePV(pv *v1.PersistentVolume, deleteFn func(pv *v1.PersistentVolume) error) error {
 	if pv.Spec.Local == nil {
 		return fmt.Errorf("Unsupported volume type")
 	}
@@ -142,6 +142,7 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
 	if err != nil {
 		return err
 	}
+
 	volMode, err := d.getVolMode(pv)
 	if err != nil {
 		return fmt.Errorf("failed to get volume mode of path %q: %v", mountPath, err)
@@ -161,6 +162,11 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
 
 	switch state {
 	case CSSucceeded:
+		klog.Errorf("%v", state)
+		err := deleteFn(pv)
+		if err != nil {
+			klog.Warningf("callback function exited with an error: %v", err)
+		}
 		// Found a completed cleaning entry
 		klog.Infof("Deleting pv %s after successful cleanup", pv.Name)
 		if err = d.APIUtil.DeletePV(pv.Name); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/deleter/deleter.go
```diff
@@ -387,3 +391,23 @@ func (c *CleanupStatusTracker) RemoveStatus(pvName string, isJob bool) (CleanupS
 	}
 	return c.ProcTable.RemoveEntry(pvName)
 }
+
+func (d *Deleter) GetCleanPVSymlinksFunc() func(pv *v1.PersistentVolume) error {
```
rohan47, line 395: Not sure if this is the right place for this function.

rohan47:
cc @gnufied @jsafrane @rohantmp @sp98

rohantmp:
Just some context here: we use parts of sig-local-static-provisioner as a library in openshift/local-storage-operator (which used to deploy local-static-provisioner). We started using it as a library for more control of the PVs.

The relevant case here is the deleter. We populate the PV cache and other relevant bits in the deleter and run deleter.DeletePVs().  We would like to run a bit of code that actually removes the symlink when the PV is deleted after the released PV is cleaned up, and before the PV is deleted.

pkg/deleter/deleter.go
gnufied: It is not. Why are defining this function here?

pkg/deleter/deleter.go
```diff
@@ -67,7 +67,7 @@ func NewDeleter(config *common.RuntimeConfig, cleanupTracker *CleanupStatusTrack
 
 // DeletePVs will scan through all the existing PVs that are released, and cleanup and
 // delete them
-func (d *Deleter) DeletePVs() {
+func (d *Deleter) DeletePVs(deleteFn ...func(pv *v1.PersistentVolume) error) {
 	for _, pv := range d.Cache.ListPVs() {
```
gnufied, line 71: why is this a variadic argument? Are we expecting more than one callback or this is mainly because - so as we don't have to modify signature of all calls to `DeletePVs` ? If later - my vote is to fix those calls and keep just one function argument.

gnufied: Also some rudimentary unit tests would be nice.

rohantmp: There might be other people who import DeletePVs() in external projects

rohantmp: (though I doubt it)

rohan47: Will add unit tests

pkg/deleter/deleter.go
rohan47: I was thinking local-static-provisioner can provide a set of functions to be performed before PV deletion. These functions can be fetched and used by projects that are using local-static-provisioner as a library. Symlink cleanup can be one of them.

rohan47: Will move this function to LSO code base

rohan47: https://github.com/openshift/local-storage-operator/pull/254

pkg/deleter/deleter.go
```diff
@@ -67,7 +67,7 @@ func NewDeleter(config *common.RuntimeConfig, cleanupTracker *CleanupStatusTrack
 
 // DeletePVs will scan through all the existing PVs that are released, and cleanup and
 // delete them
-func (d *Deleter) DeletePVs() {
+func (d *Deleter) DeletePVs(deleteFn func(pv *v1.PersistentVolume) error) {
```
rohan47, line 70: Need a better name for deleteFn

pkg/deleter/deleter.go
```diff
@@ -212,6 +218,12 @@ func (d *Deleter) deletePV(pv *v1.PersistentVolume) error {
 	return d.runProcess(pv, volMode, mountPath, config)
 }
 
+func GetDefaultDeleteFn() func(pv *v1.PersistentVolume) error {
```
rohan47, line 221: Do we use an empty function?

rohantmp: yes, if we're not using variadic.

pkg/deleter/deleter.go
rohantmp: postCleanupHook postCleanupFn?

rohan47: sounds good

pkg/deleter/deleter.go
rohantmp: I don't think we need a function to return a function here. We could just have a func `PlaceholderFunc` or `DoNothingFunc` or even pass an anonymous func here no?

PR #265, opened 2021-09-29
Support for restarting main sync loop on ConfigMap change

### LINKED ISSUE #302: Consider supporting restart of the main sync loop on ConfigMap change without restarting container/recreating pod

Currently [provisioner does not reload configuration automatically](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/upgrading.md#how-to-update) and a manual delete and recreate of the Pod is needed to pick up the updated configuration. This can lead to some inconvenience where the user may wish to update a configuration without needing an image/spec change for the provisioner.

User story:
- existing cluster had no volumes in the discovery directory yet
- we deployed a configuration with the provisioner and realized we forgot to set the `fsType` field for a particular storage class.
- since there is no image version bump needed but because there is no hot-reload support, we needed to manually kill the pod and recreate in order to pick up the update config change.

Provisioner should support hot-reload of the ConfigMap and restart its internal sync loop including the PV informer and clean up job controller when it sees that the configuration on-disk is different than what it has in memory.

- Delete the Pod and recreate, this makes operating more difficult and manual.

yibozhuang:
**What type of PR is this?**
/kind feature

This change adds support for restarting the main sync loop
when a config change has been detected on disk from change
in the ConfigMap.

The implementation adds a ConfigWatcher go routine which will
check the if the config has been updated by loading from disk
and comparing with what is currently used. If change is detected,
it will signal to the main sync loop to restart with the updated
config. This includes restarting the informer and the job controller
if the configuration specifies use Job to clean block devices.
This change will enhance the deployment and rollout story
for the provisioner as updates to the ConfigMap will be picked up
automatically by the provisioner without needing to explicitly
restart the pod itself.

Signed-off By: Yibo Zhuang <yibzhuang@gmail.com>

Fixes #302 

**Release note**:
```
Support for watching ConfigMap changes and restarting main sync loop including informer and job controller (if specified)
```

### DIFF
--- a/cmd/local-volume-provisioner/main.go
+++ b/cmd/local-volume-provisioner/main.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/metrics"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/metrics/collectors"
+	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/watcher"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -45,6 +46,7 @@ var (
 	optListenAddress string
 	optMetricsPath   string
 	discoveryPeriod  time.Duration
+	configSyncPeriod time.Duration
 )
 
 func main() {
@@ -53,6 +55,7 @@ func main() {
 	flag.StringVar(&optListenAddress, "listen-address", ":8080", "address on which to expose metrics and readiness status")
 	flag.StringVar(&optMetricsPath, "metrics-path", "/metrics", "path under which to expose metrics")
 	flag.DurationVar(&discoveryPeriod, "discovery-period", 10*time.Second, "the period for local volume discovery")
+	flag.DurationVar(&configSyncPeriod, "config-sync-period", 5*time.Second, "the period to check if there has been any config changes")
 	flag.Parse()
 	flag.Set("logtostderr", "true")
 
@@ -85,22 +88,16 @@ func main() {
 	client := common.SetupClient()
 	node := getNode(client, nodeName)
 
+	configUpdate := make(chan common.ProvisionerConfiguration)
+	defer close(configUpdate)
+
+	configWatcher := watcher.NewConfigWatcher(common.ProvisionerConfigPath, configSyncPeriod, provisionerConfig)
+	klog.Info("Starting config watcher\n")
+	go configWatcher.Run(configUpdate)
+
 	klog.Info("Starting controller\n")
 	procTable := deleter.NewProcTable()
-
-	go controller.StartLocalController(client, procTable, discoveryPeriod, &common.UserConfig{
-		Node:              node,
-		DiscoveryMap:      provisionerConfig.StorageClassConfig,
-		NodeLabelsForPV:   provisionerConfig.NodeLabelsForPV,
-		UseAlphaAPI:       provisionerConfig.UseAlphaAPI,
-		UseJobForCleaning: provisionerConfig.UseJobForCleaning,
-		MinResyncPeriod:   provisionerConfig.MinResyncPeriod,
-		UseNodeNameOnly:   provisionerConfig.UseNodeNameOnly,
-		Namespace:         namespace,
-		JobContainerImage: jobImage,
-		LabelsForPV:       provisionerConfig.LabelsForPV,
-		SetPVOwnerRef:     provisionerConfig.SetPVOwnerRef,
-	})
+	go controller.RunLocalController(configUpdate, client, procTable, discoveryPeriod, node, namespace, jobImage, provisionerConfig)
 
 	klog.Infof("Starting metrics server at %s\n", optListenAddress)
 	prometheus.MustRegister([]prometheus.Collector{
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -42,9 +42,56 @@ import (
 	"k8s.io/utils/mount"
 )
 
+// signal represents an indication to from client to terminate a service and waits for a callback
+// indicating that the service has successfully stopped.
+type signal struct {
+	closing chan chan struct{}
+}
+
+func newSignal() *signal {
+	return &signal{
+		closing: make(chan chan struct{}),
+	}
+}
+
+func (s *signal) stop() {
+	stopped := make(chan struct{})
+	s.closing <- stopped
+	<-stopped
+}
+
+func (s *signal) close() {
+	close(s.closing)
+}
+
+// RunLocalController facilitates and manages the sync loop.
+// It launches the main sync loop and if there is an updated configuration from the ConfigWatcher,
+// it will inform the main sync loop to terminate and then will launch a new sync loop with the
+// updated configuration.
+func RunLocalController(configUpdate <-chan common.ProvisionerConfiguration, client *kubernetes.Clientset, ptable deleter.ProcTable, discoveryPeriod time.Duration, node *v1.Node, namespace, jobImage string, config common.ProvisionerConfiguration) {
+	s := newSignal()
+	defer s.close()
+
+	startController := func(config common.ProvisionerConfiguration) {
+		StartLocalController(s, client, ptable, discoveryPeriod, common.UserConfigFromProvisionerConfig(node, namespace, jobImage, config))
+	}
+	go startController(config)
+
+	for {
+		select {
+		case newConfig := <-configUpdate:
+			s.stop()
+			go startController(newConfig)
+		}
+	}
+}
+
 // StartLocalController starts the sync loop for the local PV discovery and deleter
-func StartLocalController(client *kubernetes.Clientset, ptable deleter.ProcTable, discoveryPeriod time.Duration, config *common.UserConfig) error {
+func StartLocalController(signal *signal, client *kubernetes.Clientset, ptable deleter.ProcTable, discoveryPeriod time.Duration, config *common.UserConfig) {
 	klog.Info("Initializing volume cache\n")
+
+	informerStopChan := make(chan struct{})
+	jobControllerStopChan := make(chan struct{})
 
 	var provisionerName string
 	if config.UseNodeNameOnly {
@@ -102,7 +149,7 @@ func StartLocalController(client *kubernetes.Clientset, ptable deleter.ProcTable
 	deleter := deleter.NewDeleter(runtimeConfig, cleanupTracker)
 
 	// Start informers after all event listeners are registered.
-	runtimeConfig.InformerFactory.Start(wait.NeverStop)
+	runtimeConfig.InformerFactory.Start(informerStopChan)
 	// Wait for all started informers' cache were synced.
 	for v, synced := range runtimeConfig.InformerFactory.WaitForCacheSync(wait.NeverStop) {
 		if !synced {
@@ -111,12 +158,23 @@ func StartLocalController(client *kubernetes.Clientset, ptable deleter.ProcTable
 	}
 	// Run controller logic.
 	if jobController != nil {
-		go jobController.Run(wait.NeverStop)
+		go jobController.Run(jobControllerStopChan)
 	}
 	klog.Info("Controller started\n")
 	for {
-		deleter.DeletePVs()
-		discoverer.DiscoverLocalVolumes()
-		time.Sleep(discoveryPeriod)
+		select {
+		case stopped := <-signal.closing:
+			close(informerStopChan)
+			if jobController != nil {
+				close(jobControllerStopChan)
+			}
+			stopped <- struct{}{}
+			klog.Info("Controller stopped\n")
+			return
+		default:
+			deleter.DeletePVs()
+			discoverer.DiscoverLocalVolumes()
+			time.Sleep(discoveryPeriod)
+		}
 	}
 }
--- a/pkg/util/volume_util_unsupported.go
+++ b/pkg/util/volume_util_unsupported.go
@@ -35,7 +35,7 @@ func NewVolumeUtil() (VolumeUtil, error) {
 // GetFsCapacityByte returns capacity in bytes about a mounted filesystem.
 // fullPath is the pathname of any file within the mounted filesystem. Capacity
 // returned here is total capacity.
-func (u *volumeUtil) GetFsCapacityByte(hostPath, mountPath string, mountPointMap map[string]interface{}) (int64, error) {
+func (u *volumeUtil) GetFsCapacityByte(hostPath, mountPath string) (int64, error) {
 	return 0, fmt.Errorf("GetFsCapacityByte is unsupported in this build")
 }
 
@@ -50,7 +50,7 @@ func (u *volumeUtil) IsBlock(fullPath string) (bool, error) {
 	return false, fmt.Errorf("IsBlock is unsupported in this build")
 }
 
-func (u *volumeUtil) IsLikelyMountPoint(hostPath, mountPath, file string) (bool, error) {
+func (u *volumeUtil) IsLikelyMountPoint(hostPath, mountPath string, mountPointMap map[string]interface{}) (bool, error) {
 	return false, fmt.Errorf("IsLikelyMountPoint is unsupported in this build")
 }
 
--- a/docs/faqs.md
+++ b/docs/faqs.md
@@ -4,10 +4,10 @@
 
 - [Why it's best to use UUID of disk in mount point path](#why-its-best-to-use-uuid-of-disk-in-mount-point-path)
 - [Why I need to bind mount normal directories to create PVs for them](#why-i-need-to-bind-mount-normal-directories-to-create-pvs-for-them)
-- [I updated provisioner configuration but volumes are not discovered](#i-updated-provisioner-configuration-but-volumes-are-not-discovered)
 - [I bind mounted a directory into sub-directory of discovery directory, but no PVs created](#i-bind-mounted-a-directory-into-sub-directory-of-discovery-directory-but-no-pvs-created)
 - [Failed to start when docker --init flag is enabled.](#failed-to-start-when-docker---init-flag-is-enabled)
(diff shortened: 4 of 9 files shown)


mauriciopoppe (commented):
Some comments:

- What fields of the config should be considered mutable for a runtime update?
- If there are updates in the discovery directories, what happens with the existing PVs, are they cleaned up? What happens with PVs that are bound to workloads that are no longer accessible because the discovery directory changed?
- Consider adding an e2e test, consider the cases above

It'd be great to start a discussion over the questions above before implementing more changes

yibozhuang:
> * What fields of the config should be considered mutable for a runtime update?

In this particular PR any change to the config will trigger a restart of the controller. The current deployment story is that if the ConfigMap is updated, the running pod for the static provisioner will not be able to pick up this updated config until it is restarted or there is an image or spec change update.

One user story I had was that we needed to set the `fsType` config for a particular storage class and there was no image version bump needed but because there is no hot-reload support, we needed to manually kill the pod and recreate in order to pick up this config change.

I'm happy to further drill down specific fields that we'd want to have the runtime restart support if the direction becomes "we want this but would need to be very specific for the fields supported". My original intent of this change was to provide an option to allow for config changes without the need to recreate the pods.

> * If there are updates in the discovery directories, what happens with the existing PVs, are they cleaned up? What happens with PVs that are bound to workloads that are no longer accessible because the discovery directory changed?

Any existing PVs will not be cleaned up or changed. The provisioner will just watch on a different directory when the controller restarts. It would be up to the operator/user to handle the clean up of the existing PVs. This would be the same behavior today without this change. For e.g. if I want to change the discovery directory for the ConfigMap, then I would need a migration process for the existing PVs before restarting the provisioner to pick up the new config.

My understanding is that the previous directories would still be kept around and needed to be cleaned up just like on a config change today without this change.

> What happens with PVs that are bound to workloads that are no longer accessible because the discovery directory changed?

This is not in scope for this change. Even without this change, if I have a config change for the discovery directory and manually restart the provisioner, I would still need to worry about the migration story.

> * Consider adding an e2e test, consider the cases above

I would be more than happy to add the e2e test once we agree on the scope of the change being "restart controller on any config changes" or "only allow this for a limited number of fields".

> It'd be great to start a discussion over the questions above before implementing more changes

Sounds great!

yibozhuang:
> One possible step to do if the container is requested to be terminated with SIGTERM or if LVP detects a config change:
> 
> * go over the list of cached PVs references
>   
>   * if the PV is unbounded (cleanup the PV contents, remove the PV), I think this is not implemented.
>   * if the PV is bounded (leave the PV as it is, I think that if LVP is restarted the PV would join the cache and its contents wouldn't be cleaned up), I think that this is the default behavior for unbounded/bounded PVs i.e. do nothing on SIGTERM.
 
Correct currently the LVP will not touch existing PVs on a restart with a config change. I guess I'm not sure why we would want to have the clean up of unbounded PV be something that LVP should handle on SIGTERM or config change? The existing behavior of not doing anything makes more sense.

Also, for CSI external provisioner, there is no such automation as well and the CSI vendor will need to ensure that changes are backwards compatible or a manual clean up of existing provisioned PV will need to be done.

> Let's say that you change the `fstype` field in the config, if the above isn't also run then the existing PVs would continue to exist as it is, as far as I know LVP isn't going to update the fstype of the PVs that are already created, I haven't tried this scenario but you'd still need to manually remove the PVs, LVP would rediscover them in the discovery directory and create a PV with the updated fstype.

Correct and I think we want to continue having LVP behave this way.

> > if I have a config change for the discovery directory and manually restart the provisioner, I would still need to worry about the migration story.
> 
> True, I think that at least you'd have time to think about the manual cleanup somewhere in between stopping LVP, changing the config and starting LVP again, I'm a concerned about what would happen if this is automated.

I think in this case the config update would still be controlled by the user, this change would only handle the restart. So users would still have time to think about the manual clean up and when ready, then proceed to update the config.

I'd be happy to discuss in Slack as well or have a short sync with you to make sure we address all the current concerns. Thanks!

mauriciopoppe:
Thanks for your replies and sorry for the delayed response, I forgot about this thread

> Also, for CSI external provisioner, there is no such automation as well and the CSI vendor will need to ensure that changes 
are backwards compatible or a manual clean up of existing provisioned PV will need to be done.

Got it, thanks for checking what other sidecars do

> I'd be happy to discuss in Slack as well or have a short sync with you to make sure we address all the current concerns. Thanks!

Sure, that sounds good to me, I still think that we should list the possible scenarios (e.g. a matrix of: config map field changes, PV bound/unbound, outcome) in a doc that can be shared with other people that can give more feedback, this matrix can be later used in the docs for this new feature too.

yibozhuang:
> Sure, that sounds good to me, I still think that we should list the possible scenarios (e.g. a matrix of: config map field changes, PV bound/unbound, outcome) in a doc that can be shared with other people that can give more feedback, this matrix can be later used in the docs for this new feature too.

Hi @mauriciopoppe apologies for not following up, been busy with some other priorities.

Thanks a lot for your suggestions. I have updated the `faqs.md`, `provisioner.md`, and `upgrading.md` under `docs` highlighting this feature change.

Also filed #302 tracking this.

In `provisioner.md` I created a section section
```
Updating configuration without restarting provisioner
```
showing a matrix of each field in the configuration and its effect for existing PVs and PVs to be provisioned (i.e. new).

Can you give it another review and let me know if there is any other info that is needed and/or if you have any feedback?

Thanks again!

pkg/controller/controller.go
```diff
@@ -42,10 +42,45 @@ import (
...
+
+	go StartLocalController(stopChan, stopped, client, ptable, discoveryPeriod, common.UserConfigFromProvisionerConfig(node, namespace, jobImage, config))
+
+	for {
+		select {
+		case newConfig := <-restartController:
+			stopChan <- struct{}{}
+			<-stopped
+			go StartLocalController(stopChan, stopped, client, ptable, discoveryPeriod, common.UserConfigFromProvisionerConfig(node, namespace, jobImage, newConfig))
+		default:
```
mauriciopoppe, line 66: I think this method should block until we get a new config, do we need the default case?

pkg/controller/controller.go
```diff
@@ -42,10 +42,45 @@ import (
...
+func RunLocalController(resync time.Duration, restartController <-chan common.ProvisionerConfiguration, client *kubernetes.Clientset, ptable deleter.ProcTable, discoveryPeriod time.Duration, node *v1.Node, namespace, jobImage string, config common.ProvisionerConfiguration) {
+	stopped := make(chan struct{})
+	stopChan := make(chan struct{})
+
+	defer func() {
+		close(stopped)
+		close(stopChan)
+	}()
+
+	go StartLocalController(stopChan, stopped, client, ptable, discoveryPeriod, common.UserConfigFromProvisionerConfig(node, namespace, jobImage, config))
```
mauriciopoppe, line 58: To avoid code duplication:

```
startController := func() {
  StartLocalController(stopChan, stopped, client, ptable, discoveryPeriod, common.UserConfigFromProvisionerConfig(node, namespace, jobImage, config))
}
go startController()

/// 

    		case newConfig := <-restartController:
			stopChan <- struct{}{}
			<-stopped
			go startController()
```

pkg/controller/controller.go
```diff
@@ -111,12 +146,21 @@ func StartLocalController(client *kubernetes.Clientset, ptable deleter.ProcTable
 	}
 	// Run controller logic.
 	if jobController != nil {
-		go jobController.Run(wait.NeverStop)
+		go jobController.Run(jobControllerStopChan)
```
mauriciopoppe, line 149: what would happen if the config is reloaded while a cleanup job is running?

pkg/watcher/watcher.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Run will start running the ConfigWatcher in a loop. During each reload cycle,
+// it loads the configuration from the config file on disk and compares with the last applied
+// configuration. If there is a difference, it will send the loaded configuration to the
+// channel to restart the controller and then update its last applied configuration.
+func (cw *ConfigWatcher) Run(restartController chan<- common.ProvisionerConfiguration) {
+	for {
+		select {
+		case <-time.After(cw.resyncPeriod):
+			provisionerConfig := common.ProvisionerConfiguration{
```
mauriciopoppe, line 58: `provisionerConfig` looks like it can be moved outside the loop, also could `MinResyncPeriod` be `cw.ResyncPeriod`?

mauriciopoppe: my bad I didn't see that the value of `MinResyncPeriod` is different e.g 5s vs 5m

cmd/local-volume-provisioner/main.go
```diff
@@ -85,22 +88,16 @@ func main() {
 	client := common.SetupClient()
 	node := getNode(client, nodeName)
 
+	restartController := make(chan common.ProvisionerConfiguration)
```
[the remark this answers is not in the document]
yibozhuang: Sounds good, renamed to `configUpdate`
Thanks for the suggestion!

pkg/controller/controller.go
yibozhuang: Sure thing, fixed.

pkg/controller/controller.go
yibozhuang: Yes you are right, no need the default case, fixed! Thanks

pkg/watcher/watcher.go
yibozhuang: Thanks! Yes, moved `provisionerConfig` to outside of the loop.

pkg/controller/controller.go
yibozhuang: Summarizing in a table

| Old Config | New Config | Behavior
| ----------  | -----------  | ---------
| Job             | Job              | clean up done once using `Job`, job controller will just restart
| Job             | ProcTable   | clean up done twice on the volume
| ProcTable  | ProcTable    | reconstruct the clean up process to clean the volume again on restart
| ProcTable   | Job             | construct a `Job` resource to finish up the clean up of the volume

mauriciopoppe: Cool, thanks for the detailed explanation

pkg/controller/controller.go
```diff
@@ -42,10 +42,46 @@ import (
 	"k8s.io/utils/mount"
 )
 
+// RunLocalController facilitates and manages the sync loop.
+// It launches the main sync loop and if there is an updated configuration from the ConfigWatcher,
+// it will inform the main sync loop to terminate and then will launch a new sync loop with the
+// updated configuration.
+func RunLocalController(configUpdate <-chan common.ProvisionerConfiguration, client *kubernetes.Clientset, ptable deleter.ProcTable, discoveryPeriod time.Duration, node *v1.Node, namespace, jobImage string, config common.ProvisionerConfiguration) {
+	stopped := make(chan struct{})
```
mauriciopoppe, line 50: This pattern of sending a signal and waiting for a callback can be done with a `chan chan err` too as described here: https://go.dev/talks/2013/advconc.slide#25

pkg/watcher/watcher.go
```diff
@@ -0,0 +1,76 @@
...
+		lastAppliedConfig: config,
+	}
+}
+
+// Run will start running the ConfigWatcher in a loop. During each reload cycle,
+// it loads the configuration from the config file on disk and compares with the last applied
+// configuration. If there is a difference, it will send the loaded configuration to the
+// channel indicating restart controller and then update its last applied configuration.
+func (cw *ConfigWatcher) Run(configUpdate chan<- common.ProvisionerConfiguration) {
+	provisionerConfig := common.ProvisionerConfiguration{
```
mauriciopoppe, line 55: one nit is that there's duplicated code in main.go to load the config, this file should probably be renamed to watcher/config.go, wdyt?

mauriciopoppe:
Adding an item for this in https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/300

pkg/watcher/watcher.go
yibozhuang: Sure sounds good, changed file name to `watcher/config.go`

pkg/controller/controller.go
yibozhuang: Thanks for the suggestion. Created a new struct called `signal` with `closing chan chan struct{}`

```go
// signal represents an indication to from client to terminate a service and waits for a callback
// indicating that the service has successfully stopped.
type signal struct {
	closing chan chan struct{}
}

func newSignal() *signal {
	return &signal{
		closing: make(chan chan struct{}),
	}
}

func (s *signal) stop() {
	stopped := make(chan struct{})
	s.closing <- stopped
	<-stopped
}

func (s *signal) close() {
	close(s.closing)
}
```

and have `StartLocalController` take in the `*signal` rather than two channels.

This does make the code cleaner, thanks for the suggestion.

pkg/util/volume_util_unsupported.go
```diff
@@ -35,7 +35,7 @@ func NewVolumeUtil() (VolumeUtil, error) {
 // GetFsCapacityByte returns capacity in bytes about a mounted filesystem.
 // fullPath is the pathname of any file within the mounted filesystem. Capacity
 // returned here is total capacity.
-func (u *volumeUtil) GetFsCapacityByte(hostPath, mountPath string, mountPointMap map[string]interface{}) (int64, error) {
+func (u *volumeUtil) GetFsCapacityByte(hostPath, mountPath string) (int64, error) {
```
mauriciopoppe, line 38: thanks for catching these issues!

mauriciopoppe (commented):
/lgtm
/assign @msau42

msau42:
IIUC, this PR only restarts the sync loop and not the entire process right? Can you update the PR description, release note, and documentation to make sure that's correct?

yibozhuang:
> IIUC, this PR only restarts the sync loop and not the entire process right? Can you update the PR description, release note, and documentation to make sure that's correct?

@msau42 Thanks for the feedback

Updated issue description, PR description, release note, and the documentation indicating that this will restart the main sync loop including the informer and job controller (if specified using Job for block device clean up).

mauriciopoppe:
Question by @msau42, are we going to have problems with the cached state?

In the implementation there are a few threads, M is the main thread, LC is the local controller thread (spawned from M), JC is the job controller thread (spawned from LC), CW is the config watch thread (spawned from M).

As soon as CW detects a config update it sends the new config through a signal sent to LC through `configUpdate`, LC will stop its spawned controllers (JC) before restarting itself with the new config.

I was trying to analyze the cached state in LC and JC, the loop in LC is:

```go
	for {
		select {
		case stopped := <-signal.closing:
			close(informerStopChan)
			if jobController != nil {
				close(jobControllerStopChan)
			}
			stopped <- struct{}{}
			klog.Info("Controller stopped\n")
			return
		default:
			deleter.DeletePVs()
			discoverer.DiscoverLocalVolumes()
			time.Sleep(discoveryPeriod)
		}
	}
```

LC will select between stopping itself or running `deleter.DeletePVs`, `deleter.DeletePVs` is sync with respect to the data structures it owns and if a signal is attempted to be sent while `deleter.DeletePVs` is running then the caller will just block until LC gets to the next run of the select where the signal will be received, I think we're safe in the deleter state.

JC does this on its run loop:

```
func (c *jobController) Run(stopCh <-chan struct{}) {

	// make sure the work queue is shutdown which will trigger workers to end
	defer c.queue.ShutDown()

	klog.Infof("Starting Job controller")
	defer klog.Infof("Shutting down Job controller")

	// runWorker will loop until "something bad" happens.  The .Until will
	// then rekick the worker after one second
	wait.Until(c.runWorker, time.Second, stopCh)
}

func (c *jobController) runWorker() {
	for c.processNextItem() {
	}
}
```

if `c.processNextItem()` is running while a signal is received in `stopCh` it'll make the caller block, `wait.Until` executes decides whether to either accept an incoming signal or run `runWorker` so it'll exit in the next loop of `wait.Until` (https://github.com/kubernetes/apimachinery/blob/v0.23.6/pkg/util/wait/wait.go#L140), when JC is restarted we'll run `func (c *jobController) Run` again because of the previous call to `c.queue.ShutDown()` I think that new queue items wouldn't be added? https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/dab622fe70/vendor/k8s.io/client-go/util/workqueue/queue.go#L123 I'd check that @yibozhuang and probably move the queue creation to `Run`

Other local state includes an informer for jobs that's kept across calls to `func (c *jobController) Run`, that should be ok to be kept across JC restarts.

yibozhuang:
> when JC is restarted we'll run `func (c *jobController) Run` again because of the previous call to `c.queue.ShutDown()` I think that new queue items wouldn't be added? https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/dab622fe70/vendor/k8s.io/client-go/util/workqueue/queue.go#L123 I'd check that @yibozhuang and probably move the queue creation to `Run`

Just so I understand: we are worried that when JC restarts, the previous items that were in the controller queue would be lost? Am I understanding the above statement correctly?

If so, then I don't think it's an issue because the entire LC is going to be restarted (i.e. `StartLocalController()` is going to be called again after receiving the updated config) so that would call

```go
        var jobController deleter.JobController
	if runtimeConfig.UseJobForCleaning {
		labels := map[string]string{common.NodeNameLabel: config.Node.Name}
		jobController, err = deleter.NewJobController(labels, runtimeConfig)
		if err != nil {
			klog.Fatalf("Error initializing jobController: %v", err)
		}
		klog.Infof("Enabling Jobs based cleaning.")
	}
```

which would recreate the job queue and the informer and once the informer cache sync, it should trigger an event for every single `Job` resource and that should allow any previous queued work to reconcile appropriately.

My understanding here is that JC behaves similar to a regular controller which should be able to handle graceful and un-graceful restarts. Similar to today even without this change, which is if I were to delete the pod and recreate then the JobController can reconcile the resources after restarting.

@mauriciopoppe @msau42 please let me know if I mis-understood your concerns above. Thanks.

mauriciopoppe:
@yibozhuang got it, my concern was about `jobController` instance which I though it was created only once and whose `Run` method was called many times, thanks for the explanation, I see that a new instance is created on every call to `StartLocalController` and therefore we'd have a new queue too.

@msau42 is there any other concerns you have?

PR #275, opened 2021-10-26
Add Windows daemonset to the helm template

mauriciopoppe:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Adds a version of the daemonset that works in Windows nodes through a different `NodeSelector`

**Which issue(s) this PR fixes**:

Fixes #270

**Release note**:
```
New daemonset that can run in Windows nodes
```

PR #276, opened 2021-10-26
Update discovery and deletion code to work in Windows nodes through CSI Proxy

### LINKED ISSUE #269: Implement methods to get disk capacity in Windows
The `util/volume_util.go` interface defines `GetBlockCapacityByte` and it's implemented in Linux only, this is a good candidate for CSI Proxy's `GetVolumeStats`

https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/1cc3374d7a46c3367014f33851b0403771b89517/pkg/util/volume_util_linux.go#L31

### LINKED ISSUE #271: Include the CSI Proxy library

mauriciopoppe:
**What type of PR is this?**
>
/kind feature

Runs the discover and cleanup process in Windows nodes through CSI Proxy, implementation details:

- A disk must be partitioned and volumes must be mounted to the same location as the linux component (in the `C:/` drive) 

## Implementation details

- The config `hostDir` and `volumeDir` are escaped to be under `C:/` when the config is read, this is a requirement for the user to do i.e. **the user must mount their volumes to the C:/ drive in the same location as in linux**
- There are a few implementation of volumeUtils for CSI Proxy v1beta and v1, I might remove the v1beta version later but during testing I used v1beta
- Volume capacity is computed with CSI Proxy (`GetVolumeIDFromTargetPath` -> `GetVolumeStats`)
- Volume deletion is implemented with CSI Proxy (`GetVolumeIDFromTargetPath` -> `FormatVolume`)

## Testing

- Create a GKE cluster with 1 linux node and 1 windows node, each node will have 2 local SSDs
- In the windows node:
  - Initialize and disk and create a partition, mount the partition to the same location as the linux node (e.g. `C:\mnt\disks\ssd0` and `C:\mnt\disks\ssd1`)
- Build and deploy LVP in the cluster, we'll have 2 pods corresponding to the daemonsets for Linux and Windows, I deployed the `helm/generated-examples/development-gke.yaml` file
- Verify that LVP can connect to CSI Proxy (v1beta preinstalled in the GKE node)
- Verify that 2 PV are created for the Windows nodes, verify that their reported capacity is correct (corresponds to CSI Proxy's `GetVolumeStats` results)
- Create a test workload with a PVC that uses the same StorageClass as the LVP setup
- Verify that one of the PVs is bound to the PVC and to the workload
- Write something in the workload's mounted filesystem, verify that the files are in the host's filesystem too
- Delete the workload and the PVC
- Verify that the deleter reformats the mounted volume before deleting the PV

Fixes #271
Fixes #269

**Release note**:
```
Update discovery and deletion code to work in Windows nodes through CSI Proxy, after a PV gets cleaned by local volume provisioner it'll reformat the backing volume with NTFS 
```

/cc @jingxu97


helm/generated_examples/baremetal-affinity.yaml
```diff
@@ -172,3 +174,128 @@ spec:
...
+    matchLabels:
+      app.kubernetes.io/name: provisioner
+      app.kubernetes.io/instance: local-static-provisioner
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: provisioner
+        app.kubernetes.io/instance: local-static-provisioner
+      annotations:
+        checksum/config: 5a2fc47afffa2a982cff53af8b99da3694484be2a032fa8f6c723580d2d7fe19
```
jingxu97, line 200: do you know what is this?

pkg/discovery/discovery.go
```diff
@@ -360,10 +362,13 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
...
-			if _, isMntPnt := mountPointMap[filePath]; isMntPnt == false {
-				discoErrors = append(discoErrors, fmt.Errorf("path %q is not an actual mountpoint", filePath))
-				continue
+			// TODO(mauriciopoppe): skipping the mountpoint verification in Windows, unlike Linux there's no /proc/mounts file to check
+			if runtime.GOOS != "windows" {
+				// Validate that this path is an actual mountpoint
+				if _, isMntPnt := mountPointMap[filePath]; isMntPnt == false {
+					discoErrors = append(discoErrors, fmt.Errorf("path %q is not an actual mountpoint", filePath))
+					continue
+				}
```
jingxu97, line 371: seems like I did something here https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/272/files#diff-10bda8021b225706da64124ae6d771ef8da346af6c535e8dc1a3cecfc7d5887fR289

but maybe it is not working well since I comment it out? Just as a reference.

pkg/util/csi_proxy_v1_windows.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return &CSIProxyV1{
+		VolumeClient: volumeClient,
+	}, nil
+}
+
+// GetAPIVersions returns the versions of the client APIs.
+func (proxy *CSIProxyV1) GetAPIVersions() string {
+	return fmt.Sprintf(
+		"API Versions Volume: %s",
```
jingxu97, line 48: what volume mean here?

helm/generated_examples/baremetal-affinity.yaml
mauriciopoppe: the template in `helm/` generates this, more docs: https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

mauriciopoppe:
/hold 
https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/275 should be merged first

pkg/discovery/discovery.go
mauriciopoppe: ok, this is a simple check to see if the location discovered is in the /proc/mounts file

This line

```
		mountPoints, err := d.RuntimeConfig.Mounter.List()
```

returns an empty array in k/mount-utils or k/k/utils/mount

Anyways, this check is only run in Linux because of that

pkg/util/csi_proxy_v1_windows.go
mauriciopoppe: it's the version of the Volume api group, this is what I also did in PD CSI to check the versions of the APIs that were used

jingxu97: oh, maybe use CSI_PROXY API version?

pkg/util/csi_proxy_windows.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// FormatVolume formats a volume identified by `volumeId`
+	FormatVolume(volumeId string) (err error)
+}
+
+// NewCSIProxy returns an instance of the CSIProxy client compatible with either v1 or v1beta
+func NewCSIProxy() (CSIProxy, error) {
+	csiProxyV1, err := NewCSIProxyV1()
+	if err == nil {
+		klog.V(4).Infof("using CSIProxyV1, %s", csiProxyV1.GetAPIVersions())
```
jingxu97, line 43: maybe use V(2) level log in this funciton

pkg/util/csi_proxy_windows.go
```diff
@@ -0,0 +1,55 @@
...
+		return csiProxyV1, nil
+	}
+	klog.V(4).Infof("failed to connect to csi-proxy v1 with error=%v, will try with v1Beta", err)
+
+	csiProxyV1Beta, err := NewCSIProxyV1Beta()
+	if err == nil {
+		klog.V(4).Infof("using CSIProxyV1Beta, %s", csiProxyV1Beta.GetAPIVersions())
+		return csiProxyV1Beta, nil
+	}
+	klog.V(4).Infof("failed to connect to csi-proxy v1beta with error=%v", err)
```
jingxu97, line 53: Use klog.errorf()?

pkg/util/volume_util_windows.go
```diff
@@ -0,0 +1,89 @@
...
+	return totalBytes, nil
+}
+
+// DeleteContents deletes all the contents under the given directory
+func (u *volumeUtil) DeleteContents(mountPath string) error {
+	volumeID, err := u.csiProxy.GetVolumeId(mountPath)
+	if err != nil {
+		return err
+	}
+	err = u.csiProxy.FormatVolume(volumeID)
```
jingxu97, line 65: it will format with ntfs, right?
what if previous format type is not ntfs?

pkg/util/volume_util_windows.go
```diff
@@ -0,0 +1,89 @@
...
+	return 0, fmt.Errorf("GetBlockCapacityByte is unsupported in this build")
+}
+
+// IsBlock for unsupported platform returns error.
+func (u *volumeUtil) IsBlock(fullPath string) (bool, error) {
+	return false, fmt.Errorf("IsBlock is unsupported in this build")
+}
+
+func (u *volumeUtil) IsSymLink(fullPath string) (bool, error) {
+	return false, fmt.Errorf("IsBlock is unsupported in this build")
```
jingxu97, line 84: modify error message and below.

jingxu97: is IsSyLink used anywhere?

pkg/discovery/discovery.go
```diff
@@ -360,12 +361,20 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 				discoErrors = append(discoErrors, fmt.Errorf("path %q of filesystem mode cannot be used to create block volume", filePath))
 				continue
 			}
-			// Validate that this path is an actual mountpoint
-			if _, isMntPnt := mountPointMap[filePath]; isMntPnt == false {
-				discoErrors = append(discoErrors, fmt.Errorf("path %q is not an actual mountpoint", filePath))
-				continue
+			// TODO(mauriciopoppe): skipping the mountpoint verification in Windows, unlike Linux there's no /proc/mounts file to check
```
mauriciopoppe, line 364: I'm gonna address this TODO in the e2e PR

pkg/util/csi_proxy_v1_windows.go
mauriciopoppe: it's printed with the prefix CSI Proxy here: 

```
func NewCSIProxy() (CSIProxy, error) {
	csiProxyV1, err := NewCSIProxyV1()
	if err == nil {
		klog.V(2).Infof("using CSIProxyV1, %s", csiProxyV1.GetAPIVersions())
		return csiProxyV1, nil
	}
	klog.V(4).Infof("failed to connect to csi-proxy v1 with error=%v, will try with v1Beta", err)
```

pkg/util/csi_proxy_windows.go
mauriciopoppe: done

pkg/util/volume_util_windows.go
mauriciopoppe: > it will format with ntfs, right?
yes, that's the default format of the CSI Proxy VolumeFormat call

> what if previous format type is not ntfs?
I formatted a VHD with fat32 and then simulated the `deleteContents` call by doing a reformat with ntfs, it didn't complain:

```
PS C:\tmp> Get-VHD -Path $vhdPath | Get-Partition | Get-Volume | Format-Volume -Filesystem fat32 -Confirm:$false

DriveLetter FriendlyName FileSystemType DriveType HealthStatus OperationalStatus SizeRemaining       Size
----------- ------------ -------------- --------- ------------ ----------------- -------------       ----
                         FAT32          Fixed     Healthy      OK                   1003.93 MB 1003.94 MB


PS C:\tmp> Get-VHD -Path $vhdPath | Get-Partition | Get-Volume | Format-Volume -Filesystem ntfs -Confirm:$false

DriveLetter FriendlyName FileSystemType DriveType HealthStatus OperationalStatus SizeRemaining       Size
----------- ------------ -------------- --------- ------------ ----------------- -------------       ----
                         NTFS           Fixed     Healthy      OK                    991.11 MB 1007.93 MB

```

pkg/util/volume_util_windows.go
mauriciopoppe: I modified the message, it's probably used by the linux implementation and it needs to be implemented in Windows too because of the common interface between them

pkg/util/csi_proxy_windows.go
mauriciopoppe: done

mauriciopoppe:
/hold cancel
#275 was merged, please take a look again @jingxu97 :)

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
+	github.com/kubernetes-csi/csi-proxy/client v1.0.1
```
jingxu97, line 8: Use 1.0.2?

pkg/deleter/deleter.go
```diff
@@ -266,6 +267,11 @@ func (d *Deleter) cleanFilePV(pv *v1.PersistentVolume, mountPath string, config
 	klog.Infof("Deleting PV file volume %q contents at hostpath %q, mountpath %q", pv.Name, pv.Spec.Local.Path,
 		mountPath)
 
+	if runtime.GOOS == "windows" {
+		// the path to use is the one sent to CSI Proxy and must be in the context of the host
+		return d.VolUtil.DeleteContents(pv.Spec.Local.Path)
```
jingxu97, line 272: maybe add some details about this logic?

pkg/util/csi_proxy_v1beta_windows.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	volumeapi "github.com/kubernetes-csi/csi-proxy/client/api/volume/v1beta2"
+	volumeclient "github.com/kubernetes-csi/csi-proxy/client/groups/volume/v1beta2"
+)
+
+// CSIProxyV1Beta is the CSI Proxy implementation that uses the v1 API
```
jingxu97, line 30: You mean use v1beta2 API?
Also do you still need beta API support?

pkg/util/volume_util_windows.go
jingxu97: so after user delete pvc, the disk becomes empty disk with ntfs no matter what the previous fs. Maybe it is ok, but need release note about it, I think..

mauriciopoppe: release note updated

go.mod
mauriciopoppe: ack

pkg/util/csi_proxy_v1beta_windows.go
mauriciopoppe: we don't need it however it was useful to test against a GKE cluster that had v1beta still, I can remove it afterwards if it's not needed

pkg/deleter/deleter.go
mauriciopoppe: I added another sentence trying to clarify it

pkg/util/volume_util_windows.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	return totalBytes, nil
+}
+
+// DeleteContents deletes all the contents under the given directory
+func (u *volumeUtil) DeleteContents(mountPath string) error {
+	volumeID, err := u.csiProxy.GetVolumeId(mountPath)
+	if err != nil {
+		return err
+	}
```
msau42, line 64: On Linux, we support subdirectories of a filesystem that are bind mounted. Will this work on Windows or will we reformat the entire disk?

mauriciopoppe: In Windows the mount powershell commands use a volume and multiple logical volumes can be created in a disk, the creation of the volumes is left to the users and could happen during the node bootstrap, [the Google Cloud docs talks about creating a single partition with the max size](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/local-ssd#formatting_local_ssds_for_windows_server_clusters) but I believe they can do something like:

```
# format a 10GB disk into 10 partitions, execute this command 10 times
PS C:\disks> Get-Disk -Number 1 | New-Partition -Size 1073741824
PS C:\disks> Get-Disk -Number 1 | Get-Partition

   DiskPath: \\?\scsi#disk&ven_msft&prod_virtual_disk#2&1f4adffe&0&000001#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}

PartitionNumber  DriveLetter Offset                                     Size Type
---------------  ----------- ------                                     ---- ----
2                            16777216                                   1 GB Basic
3                            1090519040                                 1 GB Basic
4                            2164260864                                 1 GB Basic
5                            3238002688                                 1 GB Basic
6                            4311744512                                 1 GB Basic
7                            5385486336                                 1 GB Basic
8                            6459228160                                 1 GB Basic
9                            7532969984                                 1 GB Basic
10                           8606711808                                 1 GB Basic

PS C:\disks> Get-Disk -Number 1 | Get-Partition | Get-Volume | Format-Volume -Filesystem ntfs -Confirm:$false

DriveLetter FriendlyName FileSystemType DriveType HealthStatus OperationalStatus SizeRemaining    Size
----------- ------------ -------------- --------- ------------ ----------------- -------------    ----
                         NTFS           Fixed     Healthy      OK                   1007.15 MB 1024 MB
                         NTFS           Fixed     Healthy      OK                   1007.15 MB 1024 MB
                         NTFS           Fixed     Healthy      OK                   1007.15 MB 1024 MB
                         NTFS           Fixed     Healthy      OK                   1007.15 MB 1024 MB
                         NTFS           Fixed     Healthy      OK                   1007.15 MB 1024 MB
                         NTFS           Fixed     Healthy      OK                   1007.15 MB 1024 MB
                         NTFS           Fixed     Healthy      OK                   1007.15 MB 1024 MB
                         NTFS           Fixed     Healthy      OK                   1007.15 MB 1024 MB
                         NTFS           Fixed     Healthy      OK                   1007.15 MB 1024 MB

PS C:\disks> (Get-Disk -Number 1 | Get-Partition | Get-Volume).UniqueId
\\?\Volume{452e318a-5cde-421e-9831-b9853c521012}\
\\?\Volume{e56d239c-f207-41d8-b942-111bbde76516}\
\\?\Volume{c148a6bc-22cb-49af-9119-3549ddf302c5}\
\\?\Volume{a0c91227-1d85-4836-83a1-0dac975ef81d}\
\\?\Volume{e978bf1a-1d2f-4a0e-8b41-600a774364f3}\
\\?\Volume{5db13a97-94d4-45d3-b20d-59b20dac7c97}\
\\?\Volume{a677f928-4aa6-483b-8937-d152534f9c65}\
\\?\Volume{548eb1fa-02bb-4f15-aec9-a9e95902f331}\
\\?\Volume{d01b7d64-ce64-4a54-9add-8e9b6ec238f9}\
```

The format API in CSI Proxy receives the volume's `UniqueId` shown above and will only reformat the volume and not the entire disk so I think we can do something similar to what you say.

pkg/discovery/discovery.go
mauriciopoppe: I'm not sure if this is possible, the discovery directory contains only symlinks which when dereferenced give a volumeId like `\\?\Volume{452e318a-5cde-421e-9831-b9853c521012}\` which can be understood only in the context of the host.

I'm afraid I didn't answer your question though, I'm not very familiar with how the PV works under the hood i.e. what args will be sent to the container runtime in Windows, do you have a suggestion to test this scenario?

msau42: It sounds like in windows 2) is done first and then 1) uses a symlink into the discovery directory. In that case, should we verify here that the path is a symlink?

pkg/util/volume_util_windows.go
msau42: Ok it sounds like in Windows, we only support partitions. Can we update the docs and add a Windows section?

mauriciopoppe: sure, I created https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/278

pkg/discovery/discovery.go
mauriciopoppe: yes that's a good point, right now if I remember correctly an error is collected if the path to open is not a symlink, agreed we should check if it's a symlink to consider it as the source of a PV

mauriciopoppe: the error is collected indirectly in the call GetFsCapacityByte, the first step is to dereference the symlink and if it's not a symlink then it'll fail

```
			if runtime.GOOS == "windows" {
				// the path to use is the one sent to CSI Proxy and must be in the context of the host
				// not the container, filePath is in the context of the container
				volumePathInHost := filepath.Join(config.HostDir, file)
				capacityByte, err = d.VolUtil.GetFsCapacityByte(volumePathInHost)
```

msau42: Is the error message clear on the problem?

mauriciopoppe: I created a symlink pointing to a volume `ssd0` and a directory `fake-ssd1` in the discovery dir, in the logs I saw:

```
E1206 23:51:23.874612    1568 discovery.go:222] Failed to discover local volumes: 1 error(s) while discovering volumes: [path "c:\\mnt\\disks\\fake-ssd1" fs stats error: rpc error: code = Unknown desc = error getting the volume │
│  for the mount c:\mnt\disks\fake-ssd1, internal error error getting volume from mount. cmd: (Get-Item -Path c:\mnt\disks\fake-ssd1).Target, output: , error: <nil>] 
```

I think we have the call [`isSymlink`](https://github.com/kubernetes-csi/csi-proxy/blob/master/docs/apis/filesystem_v1.md) in CSI Proxy that we can use for a better error , adding it...

mauriciopoppe: Added a couple of new commits, the last one is using CSI Proxy's IsSymlink https://github.com/kubernetes-csi/csi-proxy/blob/4c23ef2be1fbb3607619c71f3c4289b42e05342a/pkg/os/filesystem/api.go#L93 to check if the files in the discovery dir are likely mount points

The new error looks like this:

```
1207 05:24:49.494358    3028 cache.go:55] Added pv "local-pv-20666959" to cache                                                                                                                                                    I1207 05:24:49.494939    3028 discovery.go:459] Created PV "local-pv-20666959" for volume at "c:\\mnt\\disks\\ssd0" 
E1207 05:24:49.494939    3028 discovery.go:222] Failed to discover local volumes: 1 error(s) while discovering volumes: [path "c:\\mnt\\disks\\fake-ssd1" is likely not a mount point: <nil>] 
I1207 05:24:49.508014    3028 cache.go:64] Updated pv "local-pv-20666959" to cache
```

pkg/discovery/discovery.go
```diff
@@ -361,23 +361,34 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
...
 				// the path to use is the one sent to CSI Proxy and must be in the context of the host
 				// not the container, filePath is in the context of the container
-				volumePathInHost := filepath.Join(config.HostDir, file)
-				capacityByte, err = d.VolUtil.GetFsCapacityByte(volumePathInHost)
+				volumePathInHostContext := filepath.Join(config.HostDir, file)
+
+				// in Windows a path is likely a mount point if it's a symlink pointing to a path that exists
+				isLikelyMountPoint, err := d.VolUtil.IsLikelyMountPoint(volumePathInHostContext)
+				if !isLikelyMountPoint || err != nil {
+					discoErrors = append(discoErrors, fmt.Errorf("path %q is likely not a mount point: %v", filePath, err))
```
msau42, line 372: For the error message, can we make it "is not a symlink" since in Windows mount points are a different concept? A user seeing the error message should be able to easily understand that they need to create a symlink.

pkg/discovery/discovery.go
```diff
@@ -361,23 +361,34 @@ func (d *Discoverer) discoverVolumesAtPath(class string, config common.MountConf
 				discoErrors = append(discoErrors, fmt.Errorf("path %q of filesystem mode cannot be used to create block volume", filePath))
 				continue
 			}
-			// TODO(mauriciopoppe): skipping the mountpoint verification in Windows, unlike Linux there's no /proc/mounts file to check
 			if runtime.GOOS == "windows" {
 				// the path to use is the one sent to CSI Proxy and must be in the context of the host
 				// not the container, filePath is in the context of the container
-				volumePathInHost := filepath.Join(config.HostDir, file)
-				capacityByte, err = d.VolUtil.GetFsCapacityByte(volumePathInHost)
+				volumePathInHostContext := filepath.Join(config.HostDir, file)
```
msau42, line 367: Is it worth it to make this logic part of the utils? I'm not sure if there's much benefit to creating the vol utils interfaces but still have to have OS checks in the callers.

mauriciopoppe: agreed, I moved some of the code to verify the `/proc/mounts` file to the linux VolumeUtil impl, as a result there's only a single path in the discovery code

pkg/discovery/discovery.go
mauriciopoppe: The error looks like this now:

```
I1208 06:09:06.301871    7108 discovery.go:411] Found new volume at host path "c:\\mnt\\disks\\ssd0" with capacity 1056894976, creating Local PV "local-pv-2066695
9", required volumeMode "Filesystem"
I1208 06:09:06.326949    7108 cache.go:55] Added pv "local-pv-20666959" to cache
I1208 06:09:06.327831    7108 discovery.go:445] Created PV "local-pv-20666959" for volume at "c:\\mnt\\disks\\ssd0"
E1208 06:09:06.327831    7108 discovery.go:221] Failed to discover local volumes: 1 error(s) while discovering volumes: [path "c:\\mnt\\disks\\fake-ssd1" is not a valid mount point: hostPath "c:\\mnt\\disks\\fake-ssd1" is not a symlink]
```

pkg/discovery/discovery.go
msau42: Thanks, can you also apply this pattern to the deleter too?

mauriciopoppe: sure, I updated it to receive both the hostpath & mountpath

PR #286, opened 2022-01-27
[WIP] Use CSI Proxy v2alpha1 RmdirContents and GetClosestVolumeIDFromTargetPath

mauriciopoppe:
**What type of PR is this?**
>
/kind api-change

Replaces the CSI Proxy v1 with v2alpha1 with additional APIs, the new APIs support removing the contents of directories and finding volumes by following symlinks and moving up in the hierarchy.

**Release note**:
```
Update discovery and deletion controllers to work in Windows nodes through CSI Proxy
```

/cc @jingxu97 
The last commit has the important code, other commits add files in `vendor` and modify the manifests


helm/generated_examples/baremetal-cleanbyjobs.yaml
```diff
@@ -18,7 +18,7 @@ metadata:
   name: local-static-provisioner-config
   namespace: default
   labels:
-    helm.sh/chart: provisioner-2.6.0-alpha.1
+    helm.sh/chart: provisioner-2.5.0-alpha.2
```
jingxu97, line 21: why moved to lower version?

pkg/util/csi_proxy_windows.go
```diff
@@ -35,27 +35,25 @@ type CSIProxy interface {
...
+	// GetClosestVolumeIDFromTargetPath gets the closest volume id for a given target path
+	// by following symlinks and moving up in the filesystem, if after moving up in the filesystem
+	// we get to a DriveLetter then the volume corresponding to this drive letter is returned instead.
+	GetClosestVolumeIDFromTargetPath(targetPath string) (volumeID string, err error)
 }
 
 // NewCSIProxy returns an instance of the CSIProxy client compatible with either v1 or v1beta
 func NewCSIProxy() (CSIProxy, error) {
-	csiProxyV1, err := NewCSIProxyV1()
+	csiProxyV1, err := NewCSIProxyV2()
```
jingxu97, line 52: changed the name to csiProxyV2 or csiProxyV2Alpha?

pkg/util/csi_proxy_windows.go
```diff
@@ -35,27 +35,25 @@ type CSIProxy interface {
...
 	// IsSymlink checks if the given path is a symlink
 	IsSymlink(mountPath string) (isSymlink bool, err error)
+
+	// GetClosestVolumeIDFromTargetPath gets the closest volume id for a given target path
+	// by following symlinks and moving up in the filesystem, if after moving up in the filesystem
+	// we get to a DriveLetter then the volume corresponding to this drive letter is returned instead.
+	GetClosestVolumeIDFromTargetPath(targetPath string) (volumeID string, err error)
 }
 
 // NewCSIProxy returns an instance of the CSIProxy client compatible with either v1 or v1beta
```
jingxu97, line 50: update comment?

pkg/util/csi_proxy_windows.go
mauriciopoppe: thanks, fixed

pkg/util/csi_proxy_windows.go
mauriciopoppe: thanks, fixed

helm/generated_examples/baremetal-cleanbyjobs.yaml
mauriciopoppe: we didn't release v2.5.0 which is the next minor version, I was thinking we should use v2.5.0 one instead of v2.6.0, the story around it is that tags were automatically created once the was a bump in these files.

I could release v2.6.0 and skip 2.5.0 too, then we'd no longer have the lint error in the presubmits

mauriciopoppe: after an offline conversation we changed it to v2.6.0-alpha.2, we're going to skip the release of v2.5.0 and release v2.6.0 directly

mauriciopoppe:
@andyzhangx This PR had the code to make LVP work with Windows too but it wasn't merged! I think I'll revisit the strategy now that HostProcess containers is available

mauriciopoppe:
/close

I'll revisit the strategy now that HostProcess containers is stable.

PR #287, opened 2022-01-28
Prevent cleaning released volumes more than once

dobsonj:
**What type of PR is this?**
/kind bug

There is a small time window between the deleter finding a successfully
completed cleanup job and the cache entry being removed in which
DeletePVs() may be called again. If this happens, then a second cleanup
job may be started before the cache entry is removed. This fix prevents
starting another cleanup job for a cache entry that has already been
cleaned successfully.

This issue was originally encountered in a downstream component:
https://bugzilla.redhat.com/show_bug.cgi?id=2032924
/cc @gnufied @jsafrane

**Release note**:
```
NONE
```

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -28,13 +28,17 @@ import (
 // It is periodically updated by the Populator.
 // The Deleter and Discoverer use the VolumeCache to check on created PVs
 type VolumeCache struct {
-	mutex sync.Mutex
-	pvs   map[string]*v1.PersistentVolume
+	mutex   sync.Mutex
+	pvs     map[string]*v1.PersistentVolume
+	cleaned map[string]bool
 }
 
 // NewVolumeCache creates a new PV cache object for storing PVs created by this provisioner.
 func NewVolumeCache() *VolumeCache {
-	return &VolumeCache{pvs: map[string]*v1.PersistentVolume{}}
+	return &VolumeCache{
+		pvs:     map[string]*v1.PersistentVolume{},
+		cleaned: map[string]bool{},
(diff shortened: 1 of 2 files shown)


pkg/cache/cache.go
```diff
@@ -85,6 +91,24 @@ func (cache *VolumeCache) ListPVs() []*v1.PersistentVolume {
...
+	klog.Infof("Marked pv %q as cleaned in the cache", pv.Name)
+}
+
+// SuccessfullyCleanedPV returns true if the PV was cleaned but not yet deleted
+func (cache *VolumeCache) SuccessfullyCleanedPV(pv *v1.PersistentVolume) bool {
+	cache.mutex.Lock()
+	defer cache.mutex.Unlock()
+
+	cleaned, exists := cache.cleaned[pv.Name]
+	return exists && cleaned
```
dobsonj, line 109: I think this should return `!exists || cleaned` instead.
The scenario where this makes a difference is if `DeletePVs` calls `ListPVs`, and then the cache entry is removed after that list is retrieved but before `SuccessfullyCleanedPV` is called. In that case, we should not attempt to clean the PV until `AddPV` is called again (when a new cache entry is created).

dobsonj:
This issue doesn't actually seem to affect static provisioner under normal conditions since the controller is a single threaded loop:
https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/f8de8e6f0755068dbc647a7c49168cdb200a19c3/pkg/controller/controller.go#L117-L121
Closing this PR, and we'll deal with the issue downstream.

PR #292, opened 2022-03-23
ClusterRole system:persistent-volume-provisioner replaced with a custom ClusterRole with the same contents minus permissions to access PVCs

mauriciopoppe:
**What type of PR is this?**
>
/kind cleanup

The role has the same permissions minus the PVC rules because LVP doesn't interact with the PVC objects,

Summary:
- 1st commit: add the custom ClusterRole in the e2e tests
- 2nd commit : add the custom ClusterRole in the helm chart

**Release note**:
```
ClusterRole system:persistent-volume-provisioner replaced with a custom ClusterRole with the same contents minus permissions to access PVCs
```

### DIFF
--- a/helm/generated_examples/baremetal-affinity.yaml
+++ b/helm/generated_examples/baremetal-affinity.yaml
@@ -6,7 +6,7 @@ metadata:
   name: local-static-provisioner
   namespace: default
   labels:
-    helm.sh/chart: provisioner-2.6.0-alpha.0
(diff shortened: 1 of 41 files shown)


mauriciopoppe:
/cc @msau42 
/uncc @saad-ali @wongma7 

All the e2e tests passed with the custom ClusterRole that doesn't have PVC rules

PR #301, opened 2022-04-04
Bump prometheus/client_golang to v1.11.1

dobsonj:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Bumps prometheus/client_golang to v1.11.1 to address [CVE-2022-21698](https://nvd.nist.gov/vuln/detail/CVE-2022-21698)

#### Does this PR introduce a user-facing change?

```release-note
NONE
```


vendor/modules.txt
```diff
@@ -172,7 +172,7 @@ github.com/onsi/gomega/types
 github.com/opencontainers/go-digest
 # github.com/pkg/errors v0.9.1
 github.com/pkg/errors
-# github.com/prometheus/client_golang v1.12.1
+# github.com/prometheus/client_golang v1.12.1 => github.com/prometheus/client_golang v1.11.1
```
mauriciopoppe, line 175: sorry to get back to this PR so late, I'm not sure if the prometheus version used by this library is already greater than 1.11.1, from this line it looks like it

dobsonj: > sorry to get back to this PR so late, I'm not sure if the prometheus version used by this library is already greater than 1.11.1, from this line it looks like it

Ah, you're absolutely right, sorry for the extra noise. This was originally fixed in [1.12.0](https://github.com/prometheus/client_golang/releases/tag/v1.12.0) and later backported to [1.11.1](https://github.com/prometheus/client_golang/releases/tag/v1.11.1). So there's no need to merge this PR after all.
Thanks for reviewing!

Issue #297, opened 2022-03-28
The canary & post-submit jobs are failing in test-infra


mauriciopoppe:
**What happened**:

In preparation for a new release I saw that the post submit jobs are failing in test infra:

- Canary job: https://testgrid.k8s.io/sig-storage-image-build#canary-sig-storage-local-static-provisioner-push-images
- Post submit job: https://testgrid.k8s.io/sig-storage-image-build#post-sig-storage-local-static-provisioner-push-images

mauriciopoppe:
I checked that we have 2 places where we're trying to run post-submit

- [sig-storage-local-static-provisioner](https://k8s-testgrid.appspot.com/sig-storage-local-static-provisioner)
  - the canary job is running fine: https://k8s-testgrid.appspot.com/sig-storage-local-static-provisioner#master-canary-build
  - the release job wasn't run in a long time, there are no logs or history for its latest run
- [sig-storage-image-build](https://k8s-testgrid.appspot.com/sig-storage-image-build#canary-sig-storage-local-static-provisioner-push-images) (generated by https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-sig-storage.sh)
  - the canary script uses cloudbuild but the setup is not done yet and the script is broken, I need to update this line https://github.com/kubernetes/test-infra/blob/5ec7217c27b74c9882168b4d81d8c0c8c636d371/config/jobs/image-pushing/k8s-staging-sig-storage.sh#L123
  - the release script uses cloudbuild but the setup is not done yet

Still to decide: which of these post-submit jobs should own the image pushing

- sig-storage-image-build
  - we follow what's done in CSI which is well documented and there's a lot of code already done that supports it
  - we only need a `cloudbuild.yaml` file and it'll get automatically pushed to k8s-staging-sig-storage
  - does CI every week instead of "per push builds"
-  sig-storage-local-static-provisioner
  - more manual, we do have control over the release scripts too in test-infra which unlike image-build are not generated by a script
  - does CI for every commit to master

Probably a combination would also work great, we could take the sig-storage tools built around image publishing and the custom prow jobs for commits to master

andyzhangx:
it's fixed now: https://k8s-testgrid.appspot.com/sig-storage-image-build#post-sig-storage-local-static-provisioner-push-images

PR #308, opened 2022-04-21
Increase build timeout to 2h

mauriciopoppe:
**What type of PR is this?**
>
/kind cleanup

Increases the build timeout to 1h30m, with that the sig-storage-image-build canary job should pass (right now it's timing out at 30m https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/canary-sig-storage-local-static-provisioner-push-images/1516187005245984768).

Should fix canary builds for sig-storage-image-build

**Release note**:
```release-note
NONE
```

/assign @msau42

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -4,7 +4,8 @@
 #
 #    gcloud builds submit --verbosity info --substitutions _STAGING_PROJECT=<your-project>,_PULL_BASE_REF=master [SOURCE]
 #
-timeout: 1800s
+# 5400s = 1h30m
+timeout: 5400s
 # This prevents errors if you don't use both _GIT_TAG and _PULL_BASE_REF,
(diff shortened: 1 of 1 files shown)


msau42:
2 hours is pretty long to do a build. Do we know what is taking so long?

mauriciopoppe:
I think it's because we build images for Windows too, there's a [canary postsubmit prow job](https://k8s-testgrid.appspot.com/sig-storage-local-static-provisioner#master-canary-build) that runs on every commit to master (described in detail https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/297#issuecomment-1081254219), there's a similar job whose definition comes from sig-storage-image-build which is the one that we want to enable.

In the canary job that already runs successfully the runtime is around 1h2m42s, I decreased the timeout to 1h30m

PR #321, opened 2022-05-15
fix: golang.org/x/crypto for CVE-2022-27191

andyzhangx:
**What type of PR is this?**
/kind bug

`golang.org/x/crypto/ssh` server can crash in certain conditions. This CSI driver most probably does not run any ssh server, but update the package to a safe version to be 100% sure and prevent dumb dep scanning tools from raising an issue. Using today's golang.org master, they don't use tags.

Updated using these commands:
```
go get -u golang.org/x/crypto
go mod tidy
go mod vendor
```

This PR also adds Trivy vulnerability scanner github action

Fixes #

**Release note**:
```release-note
fix: golang.org/x/crypto for CVE-2022-27191
```


ZeroMagic (approved):
/lgtm

PR #336, opened 2022-09-30
Add optional Prometheus PodMonitor, daemonset options

briend:
/kind feature

- Adds an optional Prometheus PodMonitor, which has some advantages over ServiceMonitor (no need for Service and the associated complications with Endpoints and EndpointSlices, and supports > 1000 nodes more easily) (https://github.com/prometheus-operator/prometheus-operator/pull/4635#issuecomment-1066506084)
- Allows disabling the extra DaemonSet for Windows OS (or Linux if you don't need that)
- Adds metricRelabelings to ServiceMonitor for more flexibility

I didn't bump the `appVersion` from `2.4.0` since I have not tested `2.5.0` but it seems like it should be switched

**Release note**:
```release-note
Optional Prometheus PodMonitor can support >1000 nodes more easily
```


briend:
I've opened an issue for discussing the original problem (we can't scrape metrics from more than 1000 nodes): https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/380

mauriciopoppe:
Thanks for creating the issue, I see that the k8s object introduced isn't a core kubernetes object (I see the PodMonitor CR whose CRD is defined in the Prometheus operator codebase https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/podmonitor_types.go), I think that most of the storage related components don't bundle objects that aren't part of core k8s and it's up to the cluster administrator to install the appropriate CRDs for extensions.

Adding a CRD we don't own makes it hard to maintain the project too, let's say that we want to test this then we'd need to setup a cluster with the Prometheus operator installed first before installing local-static-provisioner.

An alternative approach is to install the local-static-provisioner helm chart as a dependency of some other chart, if you wrap this chart with your own chart you can add the Prometheus operator + PodMonitoring CRs that you want.

Issue #159, opened 2019-12-05
Configuration of volume mode "block" should be completed


lz006:
In order to get local-provisioner creating pvs for raw block devices host path "/dev" needs to be mounted in daemonset to "/dev" as well.

This hasn't been described in docs nor an example is provided. 

Furthermore this might has led to error in kubespray https://github.com/kubernetes-sigs/kubespray/issues/5389

cofyc:
it's not documented, but builtin in helm chart. do you suggest to document it, can you send a PR?

Kartik494:
I would like to give a try on this.
/assign

Kartik494:
@cofyc raised a PR here https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/341 . Please take a look
Thanks!!

PR #341, opened 2022-10-18
Modify daemon set to add path /dev

Kartik494:
Fixes: #159 
change the DaemonSet to add the /dev mount the local volume provisioner.

Please also See https://github.com/kubernetes-sigs/kubespray/issues/5389


andyzhangx (approved):
/lgtm
/approve

mauriciopoppe:
We have two locations for examples that we should cleanup, tracked in https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/343

PR #349, opened 2022-10-31
Update Helm chart automation

### LINKED ISSUE #345: Release Helm chart for v2.5.0
**What happened**:
There isn't a chart published for v2.5.0.

**What you expected to happen**:
The GH action is reporting as successful but there is no chart.

**How to reproduce it**:

**Anything else we need to know?**:

**Environment**:
- CSI Driver version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

### LINKED ISSUE #350: Remove generation code for helm v2
There's code that attempts to generates manifests with Helm v2 which is deprecated, I think we should update the files in hack/upgrade-generated.sh and https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/hack/verify-generated.sh?#L40

stevehipwell:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature
> /kind flake

This PR updates the Helm chart automation.

Fixes #345
Fixes #350

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
--- a/.github/workflows/helm-chart-lint.yml
+++ b/.github/workflows/helm-chart-lint.yml
@@ -1,18 +1,43 @@
 name: Lint Helm Charts
 
-on: pull_request
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - "helm/provisioner/**"
 
 jobs:
   lint-test:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
         with:
           fetch-depth: 0
 
-      - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.1.0
+      - name: Set-up Helm
+        uses: azure/setup-helm@f382f75448129b3be48f8121b9857be18d815a82
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          version: latest
+
+      - name: Set-up Python
+        uses: actions/setup-python@13ae5bb136fac2878aff31522b9efb785519f984
+        with:
+          python-version: "3.x"
+
+      - name: Set-up chart-testing
+        uses: helm/chart-testing-action@afea100a513515fbd68b0e72a7bb0ae34cb62aec
 
       - name: Run chart-testing (lint)
-        run: ct lint --chart-dirs=helm --lint-conf helm/lintconf.yaml --validate-maintainers=false
+        run: ct lint --chart-dirs=helm --charts=helm/provisioner --lint-conf=helm/lintconf.yaml --check-version-increment=false --validate-maintainers=false
+
+      - name: Create Kind cluster
+        uses: helm/kind-action@9e8295d178de23cbfbd8fa16cf844eec1d773a07
+
+      - name: Run chart-testing (install)
+        run: ct install --chart-dirs=helm --charts=helm/provisioner
--- a/helm/provisioner/Chart.yaml
+++ b/helm/provisioner/Chart.yaml
@@ -1,16 +1,14 @@
-apiVersion: v1
-name: provisioner
-description: local provisioner chart
+apiVersion: v2
+name: local-volume-static-provisioner
+description: Helm chart for the SIG Storage Local Volume Static Provisioner.
+type: application
+version: 1.0.0
+appVersion: 2.5.0
 keywords:
-- storage
-- local
-
-# This is the chart version. This version number should be incremented each time you make changes
-# to the chart and its templates, including the app version.
-# Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.6.0-alpha.1
-
-# This is the version number of the application being deployed. This version number should be
-# incremented each time you make changes to the application. Versions are not expected to
-# follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 2.4.0
+  - storage
+  - local
+  - volume
+home: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/
+icon: https://avatars.githubusercontent.com/u/36015203?s=200&v=4
+sources:
+  - https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -2,19 +2,20 @@ name: Release Helm Charts
 
 on:
   push:
-    tags:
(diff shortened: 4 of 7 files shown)


stevehipwell:
@mauriciopoppe this PR should fix the helm chart process, I've split the chart and app versions but I can re-align them if you want but I'd advise against it. I chose `v1.0.0` for the chart version but again I can change this. The chart will lint in any PR when it is modified. Only change _Chart.yaml_ for a release, changing this file on the `master` branch will result in a chart release.

.github/dependabot.yaml
```diff
@@ -0,0 +1,6 @@
+version: 2
```
mauriciopoppe, line 1: This is cool and it was proposed for other CSI repos too, I think this should come on its own PR

.github/workflows/helm-chart-lint.yml
```diff
@@ -1,18 +1,43 @@
...
 jobs:
   lint-test:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
```
mauriciopoppe, line 18: were there issues pinning a tagged release?

.github/workflows/helm-chart-lint.yml
```diff
@@ -1,18 +1,43 @@
...
         with:
           fetch-depth: 0
 
-      - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.1.0
+      - name: Set-up Helm
+        uses: azure/setup-helm@f382f75448129b3be48f8121b9857be18d815a82
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
mauriciopoppe, line 25: This requires setting up a secret through the Github UI, if that's true whose secret should we use?

helm/provisioner/Chart.yaml
```diff
@@ -1,16 +1,14 @@
-apiVersion: v1
-name: provisioner
-description: local provisioner chart
+apiVersion: v2
```
mauriciopoppe, line 1: it looks like you're making it compatible with helm3, is the presubmit Github action job, are there any other changes in the this bump? would it be tested through the kind cluster that you're trying to setup in a presubmit job?

.github/dependabot.yaml
mauriciopoppe: Previously we didn't add similar stuff like this because of maintainability, it's easy to add things and make them work but if something fails e.g. if there's a dependency bump or a vuln notification then someone has to review that too and maintain this .github/<>.yaml files too. I think it was hard to find a volunteer to do this.

stevehipwell: I don't think `v1` is still supported. Also for a chart which doesn't exist yet I can't see a valid reason to use the old version?

I'm not quite sure of the context for the second question?

.github/workflows/helm-chart-lint.yml
stevehipwell: Tags can be moved, so shouldn't be used. Using commits directly in combination with Dependabot to generate the PRs (these show the versions & change log) is the recommended way to work with GH Actions for a secure supply chain.

.github/workflows/helm-chart-lint.yml
stevehipwell: This secret is provided by GH Actions.

.github/dependabot.yaml
mauriciopoppe: is the last comment on the right thread? I think it should be in the Chart.yaml thread

helm/provisioner/Chart.yaml
mauriciopoppe: More context: 

```
      - name: Create Kind cluster
        uses: helm/kind-action@9e8295d178de23cbfbd8fa16cf844eec1d773a07
```

.github/dependabot.yaml
stevehipwell: > Previously we didn't add similar stuff like this because of maintainability, it's easy to add things and make them work but if something fails e.g. if there's a dependency bump or a vuln notification then someone has to review that too and maintain this .github/<>.yaml files too. I think it was hard to find a volunteer to do this.

This bit?

mauriciopoppe: My bad, I maintain a few repos in kubernetes-csi but I forget that this repo is in kubernetes-sigs and doesn't have the same requirements as kubernetes-csi

helm/provisioner/Chart.yaml
mauriciopoppe: yeah makes sense to catch bugs in a kind cluster too. Thanks

Kartik494:
@stevehipwell it's seems like job is failing while using `apiversion v2` 
```
==== HELM v2====
Error: apiVersion 'v2' is not valid. The value must be "v1"
```

stevehipwell:
@Kartik494 how come there is a constrain on the chart API version? Could you point me at the docs for this?

stevehipwell:
@mauriciopoppe but why is it validating for Helm v2 which is long past EOL?

https://helm.sh/blog/helm-v2-deprecation-timeline/

mauriciopoppe:
yes, I think it's fine if you comment the call in verify-generated.sh, tracking the helm v2 removal in https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/350

stevehipwell:
@mauriciopoppe I've removed all of the Helm manifest testing as the implementation is incredibly brittle and really only serves the purpose of adding an number of extra steps each time the Helm chart is changed without actually testing anything. This is because the outcomes need to be regenerated for each chart change meaning it could be inspected once if someone is actually doing it otherwise it's just being regenerated and blindly run as if it's actually proving anything even though it will implicitly pass if it's been regenerated or fail if it hasn't.

I could put them back and regenerate the outputs from the latest chart version but that doesn't sound like a productive use of anyone's time; if that kind of testing is required someone would need to spend the (significant) time to write actual unit tests or e2e tests.

stevehipwell:
@mauriciopoppe it looks like the `gh-pages` branch wasn't manually un-protected, if you do that and delete the chart release/tag you should be able to re-run the failed action.

mauriciopoppe:
@stevehipwell thanks for the steps, I see that https://kubernetes-sigs.github.io/sig-storage-local-static-provisioner/index.yaml is available.

I can do

```
helm repo add sig-storage-local-static-provisioner https://kubernetes-sigs.github.io/sig-storage-local-static-provisioner
helm template --debug sig-storage-local-static-provisioner/local-volume-static-provisioner --version 1.0.0
```

But it's showing that the image is v2.4.0, I'll create another PR bumping the image version in the chart and adding this info to the README

stevehipwell:
@mauriciopoppe if you can wait until Monday I'll open a PR to link the image tag to the chart app version (which I assumed was already in place). I didn't realise that the tag was hard coded in the values.

mauriciopoppe:
I created https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/353

PR #353, opened 2022-11-19
Use latest image in daemonset, add node selector for Windows

### LINKED ISSUE #347: Helm -- nodeSelector not applied to windows daemonset
**What happened**:

Render the helm chart (2.5.1) with the following values.yaml:

```
daemonset:
  nodeSelector:
    my-tag: "true"
```

The windows daemonset contains only:

```
      nodeSelector:
        kubernetes.io/os: windows
```

but the linux daemonset contains:

```
      nodeSelector:
        kubernetes.io/os: linux

        my-tag: "true"
```

**What you expected to happen**:

Both the linux and windows daemonsets should apply the supplied nodeSelectors.

**How to reproduce it**:

```
helm template -f values.yaml test path/to/helm/chart
```

**Anything else we need to know?**:

**Environment**:
- CSI Driver version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

mauriciopoppe:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
/kind documentation
> /kind test
> /kind failing-test
> /kind feature
> /kind flake

Helm chart v1.0.0 uses registry.k8s.io/sig-storage/local-volume-provisioner:v2.5.0,  add field .Values.daemonset.nodeSelectorWindows to the helm chart.

Fixes #347

**Release note**:
```release-note
Helm chart v1.0.0 uses registry.k8s.io/sig-storage/local-volume-provisioner:v2.5.0
Add field .Values.daemonset.nodeSelectorWindows to the helm chart.
```

/cc @andyzhangx

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -110,12 +110,21 @@ helm upgrade ./helm/provisioner -f <path-to-your-values-file> <release-name>
 
 ### helm version  >= v3.0.0
 
-Install via helm template:
+Install by adding the repo as a Helm repo:
+
+```sh
+helm repo add sig-storage-local-static-provisioner https://kubernetes-sigs.github.io/sig-storage-local-static-provisioner
+helm template --debug sig-storage-local-static-provisioner/local-static-provisioner --version <version> --namespace <namespace> > local-volume-provisioner.generated.yaml
+# edit local-volume-provisioner.generated.yaml if necessary
+kubectl create -f local-volume-provisioner.generated.yaml
+```
+
+Or install by cloning the repo locally:
 
 ```console
 git clone --depth=1 https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner.git
 helm template -f <path-to-your-values-file> <release-name> --namespace <namespace> ./helm/provisioner > local-volume-provisioner.generated.yaml
-edit local-volume-provisioner.generated.yaml if necessary
+# edit local-volume-provisioner.generated.yaml if necessary
 kubectl create -f local-volume-provisioner.generated.yaml
 ```
 
--- a/helm/generated_examples/baremetal-affinity.yaml
+++ b/helm/generated_examples/baremetal-affinity.yaml
@@ -1,25 +1,25 @@
 ---
-# Source: provisioner/templates/serviceaccount.yaml
+# Source: local-static-provisioner/templates/serviceaccount.yaml
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: local-static-provisioner
   namespace: default
   labels:
-    helm.sh/chart: provisioner-2.6.0-alpha.1
-    app.kubernetes.io/name: provisioner
+    helm.sh/chart: local-static-provisioner-1.0.0
+    app.kubernetes.io/name: local-static-provisioner
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: local-static-provisioner
 ---
-# Source: provisioner/templates/configmap.yaml
+# Source: local-static-provisioner/templates/configmap.yaml
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: local-static-provisioner-config
   namespace: default
   labels:
-    helm.sh/chart: provisioner-2.6.0-alpha.1
-    app.kubernetes.io/name: provisioner
+    helm.sh/chart: local-static-provisioner-1.0.0
+    app.kubernetes.io/name: local-static-provisioner
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: local-static-provisioner
 data:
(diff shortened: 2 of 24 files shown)


stevehipwell:
@mauriciopoppe I'd suggest that this should be chart `v1.1.0` as you're bumping the app version to `v2.5.0`.

stevehipwell:
@mauriciopoppe or if you're renaming the chart just delete the release and tag from GH and publish over the top with the current `v1.0.0` version number.

mauriciopoppe:
I added a new feature to the chart so you're right about using v1.1.0

stevehipwell:
@mauriciopoppe what about the chart name change?

mauriciopoppe:
> @mauriciopoppe or if you're renaming the chart just delete the release and tag from GH and publish over the top with the current v1.0.0 version number.

Yeah, this was to keep the existing examples have the same name, I can do this too but gh-pages has already v1.0.0 so I'm not sure if I need to remove that file or if it'll be automatically updated too

mauriciopoppe:
I think no one is using the helm chart yet so I'll recreate v1.0.0 from scratch

stevehipwell:
@mauriciopoppe is changing the name for the examples worth it? Isn't the new name more correct? Either way deleting the tag/release and index.yaml from pages branch would be my suggestion.

mauriciopoppe:
It'd be nice to refer to it as local-volume-provisioner (as that's also the name we use for it internally) but the repo name is sig-storage-local-static-provisioner. With this change there are no updates in the names, in other words before the bump to Helm apiVersion v2 they were local-static-provisioner and they're still local-static-provisioner after this change.

stevehipwell:
@mauriciopoppe why not go for `local-volume-provisioner` then? It'd be nice to have a constant name (DDD benefits etc). You could then find/replace the examples to this name.

mauriciopoppe:
I wouldn't change it to avoid confusion, if the chart has the name same as the repo I think that's less confusing but that's my opinion. If we change the repo to local-volume-provisioner that'd ideal but I think it's outside the scope of this change.

I'd bring this up in sig-storage to discuss changing the name of the repo.

mauriciopoppe:
@cofyc @msau42 could you review this please? it's to make gh-pages a helm repo, I updated the installation instructions if you want more details

/assign @msau42 @cofyc

stevehipwell:
@mauriciopoppe is there any way we can get this prioritised?

msau42:
Sorry I am not very familiar with Helm. Can someone summarize the conclusion of the discussion of chart versioning and how it relates to tags?

In general, image tags should be treated as immutable and we should not republish an existing tag even if we make a mistake.

mauriciopoppe:
- The chart tag will be independent of the image tag, eg we can add a feature to the chart which is still using an existing image. This PR is a good example where I added an additional node selector field for Windows in the chart
- There’s a github action that automatically releases special named tags for the chart if there are changes to the helm/chart directory, the name of the chart is the same as the repo eg local-static-provisioner 
- On a release it also updates a repo index, the repo index is an index.yaml file in the gh-pages branch which maps a chart version to a named tag, there’s more info for end users in one of the PR files
- There were a few runs of the GitHub action already but with the name local-volume-provisioner and with an older image tag, to avoid confusion with the name I renamed the chart to local-static-provisioner and I deleted the named tag and the index.yaml file from the gh-pages branch, it’s now pristine and if we merge the PR it should run the GitHub action again

mauriciopoppe:
> In general, image tags should be treated as immutable and we should not republish an existing tag even if we make a mistake.

@msau We didn't release a new image tag, there were new named tags created from a Github Action run and these tags have the form https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/releases/tag/provisioner-2.6.0-alpha.1

Adding to the conclusions in https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/353#issuecomment-1340387629 once the PR merges the Github Action will run and create a tag like the above, it'll also modify the `gh-pages` branch adding this file index.yaml that is used by these commands:

```
# add public index.yaml served from the gh-pages branch
helm repo add sig-storage-local-static-provisioner https://kubernetes-sigs.github.io/sig-storage-local-static-provisioner

# in the helm repo there's only one named chart (local-static-provisioner), create a single manifest to apply to the cluster
helm template --debug sig-storage-local-static-provisioner/local-static-provisioner --version <version> --namespace <namespace> > local-volume-provisioner.generated.yaml

# edit local-volume-provisioner.generated.yaml if necessary
kubectl create -f local-volume-provisioner.generated.yaml
```

msau42 (commented):
There was a new helm chart example added in https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/364. You may need to rebase to update that.

mauriciopoppe:
yes, it's up to date with master which has #364 merged

mauriciopoppe:
I verified that https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/gh-pages/index.yaml was created, I also followed the README instructions and got the expected file:

```
~/go/src/sigs.k8s.io/sig-storage-local-static-provisioner chart-update *3 ?5
❯ helm repo list
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /usr/local/google/home/mauriciopoppe/.kube/config
NAME            URL
nginx-stable    https://helm.nginx.com/stable

~/go/src/sigs.k8s.io/sig-storage-local-static-provisioner chart-update *3 ?5 
❯ helm repo add sig-storage-local-static-provisioner https://kubernetes-sigs.github.io/sig-storage-local-static-provisioner
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /usr/local/google/home/mauriciopoppe/.kube/config
"sig-storage-local-static-provisioner" has been added to your repositories

~/go/src/sigs.k8s.io/sig-storage-local-static-provisioner chart-update *3 ?5         
❯ helm template --debug sig-storage-local-static-provisioner/local-static-provisioner --version v1.0.0 --namespace dummy > local-volume-provisioner.generated.yaml
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /usr/local/google/home/mauriciopoppe/.kube/config
install.go:173: [debug] Original chart version: "v1.0.0"
install.go:190: [debug] CHART PATH: /usr/local/google/home/mauriciopoppe/.cache/helm/repository/local-static-provisioner-1.0.0.tgz
```

PR #383, opened 2023-06-18
fix: CVE-2023-2431

andyzhangx:
**What type of PR is this?**
/kind bug

fix: CVE-2023-2431

Fixes #

**Release note**:
```release-note
fix: CVE-2023-2431
```


test/e2e/e2e.go
```diff
@@ -156,7 +154,6 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 var _ = ginkgo.SynchronizedAfterSuite(func() {
 	// Run on all Ginkgo nodes
 	framework.Logf("Running AfterSuite actions on all nodes")
-	framework.RunCleanupActions()
```
mauriciopoppe, line 159: could you point out the commit in the test setup in k/k where this was removed please?

test/e2e/e2e.go
```diff
@@ -116,20 +113,21 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 		}
 	}
 
-	framework.ExpectNoError(framework.WaitForAllNodesSchedulable(c, framework.TestContext.NodeSchedulableTimeout))
+	timeouts := framework.NewTimeoutContext()
```
mauriciopoppe, line 116: Thanks a lot, I learned about this NewTimeoutContext through this change

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	k8s.io/component-base v0.27.3
 	k8s.io/klog/v2 v2.90.1
 	k8s.io/kubernetes v1.27.3
+	k8s.io/pod-security-admission v0.0.0
```
mauriciopoppe, line 21: Is this the right tag? I see that the tags match OSS https://github.com/kubernetes/pod-security-admission/tags so should this be v0.27.3?

mauriciopoppe (changes requested):
Thanks for bump to ginkgo v2, ideally we should split the PR and handle the ginkgo v2 upgrade in a different PR, would it be possible to do that?

test/e2e/e2e.go
cvvz: https://github.com/kubernetes/kubernetes/commit/fc092eb1455f2eb055258a7249b16a63e3a1d7f4

go.mod
cvvz: The dependencies with "k8s.io/" prefix are replaced with v0.27.3 in go.mod, you can find in the last few lines in go.mod file.

mauriciopoppe:
> I also think keep the changes in this PR won't cause any problem, while splitting ginkgo v2 upgrade from this PR will introduce more tedious work.

Yeah it won't cause any problem now but in the past we've had to backport CVEs, if that's the case then we should backport just the CVE and not the feature (the ginkgo v2 bump is actually a feature).

cvvz:
OK. I will create a PR for bumping ginkgo v2.

cvvz:
It's meaningless to create a PR that merely bumps ginkgo v2, actually all we need to do is to upgrade the ginkgo version in go.mod. 

However, all the changes made here are actually related to new version Kubernetes test framework changes, which relies on ginkgo v2 and not backward compatible. So, we still need to upgrade Kubernetes version anyway.

go.mod
mauriciopoppe: I think it would be nice to maintain some consistence because there are other dependencies that are v0.27.3

mauriciopoppe (commented):
> However, all the changes made here are actually related to the new version Kubernetes test framework changes, which relies on ginkgo v2 and not backward compatible. So, if we are bumping ginkgo v2, it's better upgrade Kubernetes test framework together, that's what does in this PR.

Got it, yes I thought that the CVE was referring to a change in another dependency besides k8s/ginkgo v2. 

The PR looks good to me, I'd also squash the changes.

andyzhangx:
> > However, all the changes made here are actually related to the new version Kubernetes test framework changes, which relies on ginkgo v2 and not backward compatible. So, if we are bumping ginkgo v2, it's better upgrade Kubernetes test framework together, that's what does in this PR.
> 
> Got it, yes I thought that the CVE was referring to a change in another dependency besides k8s/ginkgo v2.
> 
> The PR looks good to me, I'd also squash the changes.

@mauriciopoppe the CVE would upgrade k8s lib version, and then it hit other dependency issue, so this PR upgrades the e2e test a little. I have squashed some commits.

mauriciopoppe:
Thanks for squashing some commits
/lgtm

mauriciopoppe (approved):

PR #385, opened 2023-06-28
Add controller to clean up stale PVs/PVCs when a Node is deleted

### LINKED ISSUE #202: Recover workloads using local PVs from data loss in the cloud
This is the track issue for some related problems in the cloud when the node is permanently deleted.

I'm thinking about writing a cloud controller to automate the recovery process. Here is the proposal: https://docs.google.com/document/d/1SA9epEwA3jPwibRV0ccQwJ2UfZXoeUYKyNxNegt0vn4

related issues:

- https://github.com/kubernetes/kubernetes/issues/78756
- https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/201
- https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/issues/181

justinblalock87:
**What type of PR is this?**

/kind feature

This PR allows for workloads using Local PVs (backed by ephemeral disks) to automatically recover from Node deletion. It does so by cleaning up local PVs and PVCs associated with deleted Nodes. The cleanup is configurable in that the controller will wait for a user-defined amount of time on Node deletion to start cleanup. Further, the controller will only clean up resources associated with a user-defined StorageClass to allow for this feature to be opt-in.

This feature is needed since when Nodes are deleted while having a Local PV mounted on them, Pods using the Local PV become stuck since they are tied to PVC and PV objects that reference a deleted Node. Deleting the PVC allows the Pod to get rescheduled and deleting the stale PVs reduces clutter in the apiserver.

Fixes #202 

**Release note**:
```release-note
Optional controller to automatically clean up stale PV/PVC objects when a Node is deleted
```

### DIFF
--- a/cmd/cleanup/main.go
+++ b/cmd/cleanup/main.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package main
 
 import (
--- a/cmd/cleanup/main.go
+++ b/cmd/cleanup/main.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/cmd/cleanup/main.go
+++ b/cmd/cleanup/main.go
@@ -2,13 +2,14 @@ package main
 
 import (
 	"context"
-	"flag"
 	"time"
 
+	flag "github.com/spf13/pflag"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cleanup/controller"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cleanup/deleter"
@@ -16,13 +17,13 @@ import (
 
 // Command line flags
 var (
-	kubeconfig               = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Only required when running out of cluster.")
-	masterURL                = flag.String("master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
-	resync                   = flag.Duration("resync", 10*time.Minute, "Resync interval of the controller.")
-	storageClassName         = flag.String("storageClassName", "", "Name of StorageClass to opt-in PVs and PVCs for cleanup.")
+	kubeconfig               = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Either this or kube-api-endpoint needs to be set if the provisioner is being run out of cluster.")
+	kubeApiEndpoint          = flag.String("kube-api-endpoint", "", "Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.")
+	resync                   = flag.Duration("resync", 10*time.Minute, "Duration, in minutes, of the resync interval of the controller.")
+	storageClassNames        = flag.StringSlice("storageclass-names", []string{}, "Comma separated list of names of StorageClasses to opt-in PVs and PVCs for cleanup.")
 	workerThreads            = flag.Uint("worker-threads", 10, "Number of controller worker threads.")
-	delay                    = flag.Duration("delay", 1*time.Minute, "How much time to wait after Node deletion for resource cleanup.")
-	stalePVDiscoveryInterval = flag.Duration("stalePVDiscoveryInterval", 10*time.Second, "The Deleter will look for an cleanup stale PVs periodically on this interval.")
+	pvcDeletionDelay         = flag.Duration("pvc-deletion-delay", 60*time.Second, "Duration, in seconds, to wait after Node deletion for PVC cleanup.")
+	stalePVDiscoveryInterval = flag.Duration("stale-pv-discovery-interval", 10*time.Second, "Duration, in seconds, the PV Deleter should wait between tries to clean up stale PVs.")
 )
 
 func main() {
@@ -31,7 +32,7 @@ func main() {
 
 	ctx := context.Background()
 
-	config, err := buildConfig(*kubeconfig, *masterURL)
+	config, err := buildConfig(*kubeconfig, *kubeApiEndpoint)
 	if err != nil {
 		klog.Error(err, "Error building kubeconfig")
 		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
@@ -45,16 +46,18 @@ func main() {
 
 	factory := informers.NewSharedInformerFactory(clientset, *resync)
 	pvInformer := factory.Core().V1().PersistentVolumes()
+	pvcInformer := factory.Core().V1().PersistentVolumeClaims()
 	nodeInformer := factory.Core().V1().Nodes()
 
 	cleanupController := controller.NewCleanupController(
 		clientset,
 		pvInformer,
+		pvcInformer,
 		nodeInformer,
-		*storageClassName,
-		*delay,
+		*storageClassNames,
+		*pvcDeletionDelay,
 		*stalePVDiscoveryInterval)
-	deleter := deleter.NewDeleter(clientset, pvInformer.Lister(), nodeInformer.Lister(), *storageClassName)
+	deleter := deleter.NewDeleter(clientset, pvInformer.Lister(), nodeInformer.Lister(), *storageClassNames)
 
 	factory.Start(ctx.Done())
 
@@ -68,11 +71,11 @@ func main() {
 	}
 }
 
-func buildConfig(kubeconfig string, masterURL string) (*rest.Config, error) {
+func buildConfig(kubeconfig string, kubeApiEndpoint string) (*rest.Config, error) {
 	// If kubeconfig was passed in then try to build from that
 	// since we may be out-of-cluster.
 	if kubeconfig != "" {
-		return clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
+		return clientcmd.BuildConfigFromFlags(kubeApiEndpoint, kubeconfig)
 	}
 	// Otherwise we are in-cluster.
 	return rest.InClusterConfig()
--- a/cmd/cleanup/main.go
+++ b/cmd/cleanup/main.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cleanup/controller"
+	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cleanup/deleter"
 )
 
 // Command line flags
@@ -43,17 +44,24 @@ func main() {
 	}
 
 	factory := informers.NewSharedInformerFactory(clientset, *resync)
+	pvInformer := factory.Core().V1().PersistentVolumes()
+	nodeInformer := factory.Core().V1().Nodes()
 
 	cleanupController := controller.NewCleanupController(
 		clientset,
-		factory.Core().V1().PersistentVolumes(),
-		factory.Core().V1().Nodes(),
+		pvInformer,
+		nodeInformer,
 		*storageClassName,
 		*delay,
 		*stalePVDiscoveryInterval)
+	deleter := deleter.NewDeleter(clientset, pvInformer.Lister(), nodeInformer.Lister(), *storageClassName)
 
 	factory.Start(ctx.Done())
 
+	// Start Deleter
+	go deleter.Run(ctx, *stalePVDiscoveryInterval)
+
+	// Start controller
 	if err = cleanupController.Run(ctx, int(*workerThreads)); err != nil {
 		klog.Error(err, "Error running controller")
 		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
--- a/cmd/cleanup/main.go
+++ b/cmd/cleanup/main.go
@@ -1,0 +1,71 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"time"
+
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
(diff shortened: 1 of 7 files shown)


pkg/cleanup/controller/controller.go
```diff
@@ -0,0 +1,251 @@
...
+		go wait.UntilWithContext(ctx, c.runWorker, time.Second)
+	}
+
+	klog.Info("Started workers")
+
+	// Look for stale PVs and start timers for resource cleanup
+	c.startCleanupTimersIfNeeded()
+
+	// Run Deleter and block until channel is closed
+	deleter := deleter.NewDeleter(c.client, c.pvLister, c.nodeLister, c.storageClassName)
```
tyuchn, line 123: I'd prefer making this piece of code into another Run() function in Deleter.go because decouplation give better scalability and testability.

pkg/cleanup/deleter/deleter.go
```diff
@@ -0,0 +1,108 @@
...
+	if !ok {
+		return false, fmt.Errorf("Error retrieving node")
+	}
+
+	exists, err := common.NodeExists(d.nodeLister, nodeName)
+	return !exists && err == nil, err
+}
+
+func (d *Deleter) deletePV(pvName string) error {
+	err := d.client.CoreV1().PersistentVolumes().Delete(context.TODO(), pvName, metav1.DeleteOptions{})
```
tyuchn, line 100: consider passing in the context as a variable here

pkg/cleanup/deleter/deleter.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			utilruntime.HandleError(fmt.Errorf("PV '%s' no longer exists", pvName))
```
tyuchn, line 103: I'd prefer to use klog.Errorf (the package is `"k8s.io/klog/v2"`) for logging here and everywhere.
The benefit of klog is that it can log and handle errors with additional contextual information related to the specific component or operation, which has a good integration with cloud logging.

pkg/cleanup/deleter/deleter.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	exists, err := common.NodeExists(d.nodeLister, nodeName)
+	return !exists && err == nil, err
+}
+
+func (d *Deleter) deletePV(pvName string) error {
+	err := d.client.CoreV1().PersistentVolumes().Delete(context.TODO(), pvName, metav1.DeleteOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
```
tyuchn, line 102: Consider reducing the nested level for better readability:
if err != nil && errors.IsNotFound(err) {
klog.Infof("PV %s no longer exists", pvName)
return nil
}
return err

pkg/cleanup/deleter/deleter_test.go
```diff
@@ -0,0 +1,215 @@
...
+					t.Fatalf("Unknown initalObject type: %+v", obj)
+				}
+			}
+
+			// Start test by simulating an event.
+			deleter.DeletePVs()
+
+			actions := client.Actions()
+			for i, action := range actions {
+				print(action.GetVerb())
```
tyuchn, line 142: Use t.logf for logging in unit tests

pkg/cleanup/controller/controller.go
```diff
@@ -0,0 +1,251 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cleanup/deleter"
+	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
+)
+
+type EntryPair struct {
```
tyuchn, line 28: I don't think storing pointer to k8s objects in work queue is a good practice, because you need to frequently check if the pointer is nil and you may accidentally modify the obj itself accidentally. Considering splitting into two work queue for PV and Node, and only put name or the key of the object in the queue.

- pvQueue entryQueue workqueue.RateLimitingInterface
- nodeQueue entryQueue workqueue.RateLimitingInterface

pkg/common/common_test.go
```diff
@@ -473,3 +476,163 @@ func TestGetVolumeMode(t *testing.T) {
...
+			name:             "wrong node label key",
+			pv:               withNodeAffinity(pv(), nodeName, "wrongLabel"),
+			expectedNodeName: "",
+			expectedStatus:   false,
+		},
+	}
+
+	for _, test := range tests {
+		nodeName, ok := NodeAttachedToLocalPV(test.pv)
+		if nodeName != test.expectedNodeName || ok != test.expectedStatus {
```
tyuchn, line 554: I think it's better to separate checking error and comparing result into two steps. The structure will be something like:

if gotExpected := gotExpectedError(test.name, test.expectErr, err); gotExpected != nil {
			t.Fatal(gotExpected)
		}
if nodeName != test.expectedNodeName {
t.Errorf("got XXX expected XXX...")
}

pkg/cleanup/deleter/deleter_test.go
```diff
@@ -0,0 +1,215 @@
...
+				expectedAction := test.expectedActions[i]
+				if !reflect.DeepEqual(expectedAction, action) {
+					t.Errorf("Test %q: action %d\nExpected:\n%s\ngot:\n%s", test.name, i, expectedAction, action)
+				}
+			}
+
+			if len(test.expectedActions) > len(actions) {
+				t.Errorf("Test %q: %d additional expected actions", test.name, len(test.expectedActions)-len(actions))
+				for _, a := range test.expectedActions[len(actions):] {
+					t.Logf("    %+v", a)
```
tyuchn, line 157: Can you make this log more informative?

pkg/cleanup/controller/controller.go
justinblalock87: In that case, wouldn't it be easier to just have one workqueue with the PV names and then call the common.NodeAttachedToLocalPV(pv) that I created whenever I need the associated Node?

tyuchn: Good idea! Having one work queue storing the pv names should be enough thanks to the function you created.

justinblalock87: Great thanks! Made this change

pkg/cleanup/controller/controller.go
```diff
@@ -0,0 +1,251 @@
...
+	klog.Info("Waiting for informer caches to sync")
+	// Wait for the caches to be synced before starting workers
+	if ok := cache.WaitForCacheSync(ctx.Done(), c.nodeListerSynced, c.pvListerSynced); !ok {
+		return fmt.Errorf("failed to wait for caches to sync")
+	}
+
+	klog.Infof("Starting workers, count: &d", workers)
+	// Launch workers to process
+	for i := 0; i < workers; i++ {
+		go wait.UntilWithContext(ctx, c.runWorker, time.Second)
```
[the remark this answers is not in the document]
justinblalock87: As we discussed in our meeting, it seems to be convention to put time.Second

cmd/cleanup/main.go
```diff
@@ -0,0 +1,79 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cleanup/controller"
+	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/cleanup/deleter"
+)
+
+// Command line flags
+var (
+	kubeconfig               = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Only required when running out of cluster.")
+	masterURL                = flag.String("master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
```
msau42, line 20: I see in other examples that it's one of kubeconfig or master: https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L73.

We're also moving away from the master terminology. "kube-api-endpoint" maybe would be better? Can you also add the format of the string in the help text.

cmd/cleanup/main.go
```diff
@@ -0,0 +1,79 @@
...
+	resync                   = flag.Duration("resync", 10*time.Minute, "Resync interval of the controller.")
+	storageClassName         = flag.String("storageClassName", "", "Name of StorageClass to opt-in PVs and PVCs for cleanup.")
```
msau42, line 22: Could this be a list?

cmd/cleanup/main.go
```diff
@@ -0,0 +1,79 @@
...
+	workerThreads            = flag.Uint("worker-threads", 10, "Number of controller worker threads.")
+	delay                    = flag.Duration("delay", 1*time.Minute, "How much time to wait after Node deletion for resource cleanup.")
```
msau42, line 24: I would name the parameter more specific, like "pvc-deletion-delay". Also in the help text, specify the units.

cmd/cleanup/main.go
```diff
@@ -0,0 +1,79 @@
...
+	stalePVDiscoveryInterval = flag.Duration("stalePVDiscoveryInterval", 10*time.Second, "The Deleter will look for an cleanup stale PVs periodically on this interval.")
```
msau42, line 25: be consistent with flag name style. Sepcify units in the help text.

pkg/cleanup/controller/controller.go
```diff
@@ -0,0 +1,262 @@
...
+	eventRecorder := broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: fmt.Sprintf("cleanup-controller")})
+
+	controller := &CleanupController{
+		client:           client,
+		storageClassName: storageClassName,
+		// Delayed queue with rate limiting
+		pvQueue: workqueue.NewRateLimitingQueueWithConfig(
+			workqueue.DefaultControllerRateLimiter(),
+			workqueue.RateLimitingQueueConfig{
+				Name: "stale",
```
msau42, line 66: Be more specific in naming this to describe what it does. The name of the workqueue shows up in metrics.

pkg/cleanup/controller/controller.go
```diff
@@ -0,0 +1,262 @@
...
+		// Node is back up so cleanup is not needed.
+		return nil
+	}
+
+	pvClaimRef := pv.Spec.ClaimRef
+	if pvClaimRef == nil {
+		return nil
+	}
+
+	err = c.deletePVC(ctx, pvClaimRef.Name, pvClaimRef.Namespace)
```
msau42, line 195: You probably want to validate the PVC points back to this exact PV.

pkg/cleanup/deleter/deleter.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	exists, err := common.NodeExists(d.nodeLister, nodeName)
+	return !exists && err == nil, err
+}
+
+func (d *Deleter) deletePV(ctx context.Context, pvName string) error {
+	err := d.client.CoreV1().PersistentVolumes().Delete(ctx, pvName, metav1.DeleteOptions{})
+	if err != nil && errors.IsNotFound(err) {
+		klog.Errorf("PV %q no longer exists", pvName)
```
msau42, line 115: probably more of a warning rather than an error

pkg/common/common.go
```diff
@@ -484,3 +487,41 @@ func GetVolumeMode(volUtil util.VolumeUtil, fullPath string) (v1.PersistentVolum
...
+//	        operator: In
+//	        values:
+//	        - <node1>
+func NodeAttachedToLocalPV(pv *v1.PersistentVolume) (string, bool) {
+	nodeNames := volumeUtil.GetLocalPersistentVolumeNodeNames(pv)
+	if nodeNames == nil || len(nodeNames) == 0 {
+		return "", false
+	}
+	// We assume that there should only be one matching node.
+	return nodeNames[0], true
```
msau42, line 518: I think if there are multiple node names, then it may not be safe to delete the pvc until all nodes are deleted. I this is a corner case we should not support initially.

msau42:
Is sample deployment + README going to be part of this PR?

msau42:
Also add a release note describing the feature being added

cmd/cleanup/main.go
justinblalock87: Updated to use "kube-api-endpoint" and specified format as "URL".

cmd/cleanup/main.go
justinblalock87: Changed to list of storageclass names such that if PV/PVC belong to any one of the names in the list, they are eligible for cleanup.

cmd/cleanup/main.go
justinblalock87: Updated and specified seconds as the unit, though I think that any passed in unit should work given that the flag.Duration is parsed by time.ParseDuration().

cmd/cleanup/main.go
justinblalock87: Change flag name to stale-pv-discovery-interval.

pkg/cleanup/controller/controller.go
justinblalock87: Changed to stalePVQueue.

pkg/cleanup/controller/controller.go
justinblalock87: Add check that PVC points back to this PV before deleting. Also added a unit test for this case.

pkg/common/common.go
justinblalock87: Changed to return false if there isn't exactly one node name and added unit test.

justinblalock87:
> Is sample deployment + README going to be part of this PR?

It will be in an upcoming PR.

pkg/cleanup/controller/controller.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	exists, err := common.NodeExists(c.nodeLister, nodeName)
+	return !exists && err == nil, err
+}
+
+// deletePVC deletes the PVC with the given name and namespace
+// and returns nil if the operation was successful or if the PVC doesn't exist
+func (c *CleanupController) deletePVC(ctx context.Context, pvc *v1.PersistentVolumeClaim) error {
+	err := c.client.CoreV1().PersistentVolumeClaims(pvc.Namespace).Delete(ctx, pvc.Name, metav1.DeleteOptions{})
```
msau42, line 295: To prevent a possible conflict where someone else deletes and recreates the PVC at the same time we try to, it would be good to add the UID precondition when deleting the object: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/95d9b6325bb51f83efd7f5516a28086dcefddcf6/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L719

pkg/cleanup/controller/controller.go
```diff
@@ -0,0 +1,302 @@
...
+	if nodeExists {
+		return nil
+	}
+
+	pvClaimRef := pv.Spec.ClaimRef
+	if pvClaimRef == nil {
+		return nil
+	}
+
+	pvc, err := c.pvcLister.PersistentVolumeClaims(pvClaimRef.Namespace).Get(pvClaimRef.Name)
```
msau42, line 220: I would also use UID from the `ClaimRef` to be sure it is the exact same object.

msau42:
Great work!

/approve

pkg/cleanup/controller/controller.go
justinblalock87: Done

pkg/cleanup/controller/controller.go
justinblalock87: Done

cmd/cleanup/main.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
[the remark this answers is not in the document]
justinblalock87: Updated for new files

PR #388, opened 2023-07-05
feat: add enableWindows as helm values for Windows DaemonSet creation

### LINKED ISSUE #381: Add the ability to disable the Windows daemonset
I've encountered issues with the Windows local static provisioner daemonset (related to PSPs), and I don't actually need it to be present (we won't be using Windows nodes with it). It would be great to have a way to optionally disable either one of the daemonsets in the `values.yaml` file.

jennwah:
**What type of PR is this?**
/kind feature

feat: add enableWindows as helm values for Windows DaemonSet creation 
Add feature for enabling Windows DaemonSet if specified. By default its false.

Fixes #381

**Release note**:
```release-note
feat: add enableWindows helm chart value to control the deployment of Windows manifests
```


helm/provisioner/values.yaml
```diff
@@ -60,6 +60,10 @@ common:
   #
   #labelsForPV:
   #  pv-labels: can-be-selected
+  #
+  # Indicates if Windows DaemonSet should be created (by default: true)
+  #
+  enableWindows: true
```
mauriciopoppe, line 66: Maybe the default value should be false to make Windows deployment an opt-in, unfortunately the chart was already released so I'd argue this'd be a breaking change.

@msau42 @andyzhangx thoughts? I don't know how many people are relying on the helm chart to install the Windows component, as far as I know it was only Microsoft that's using it, maybe we can make this change in a minor patch (e.g. release 1.1.0 of the helm chart).

jennwah: It looks like perhaps making `default value = true` makes most sense here, for obvious reasons (not introducing breaking changes etc) 

In this case, perhaps `disableWindows` naming would make more sense here, since by default we are providing `linux` and `windows` DaemonSet.  

What do you guys think? @mauriciopoppe @andyzhangx @msau42 
If you agree, I would change this to `disableWindows` to allow users to disable Windows if for some users who won't need Windows DaemonSets. Also, adding documentation to `sig-storage-local-static-provisioner/helm/README.md` for this new feature, to allow users to disable windows deployment.

jennwah: @msau42 @andyzhangx @mauriciopoppe What do you think guys? Is this feature necessary, considering there're users asking for disabling windows deployment for its redundancy.

mauriciopoppe: yeah I think we can add it however we need to be ok whether it should be enabled by default

jennwah: @andyzhangx @msau42 please give your valuable opinions here for this feature addition. 

Anything would happen if we don't enableWindows by default? How many users would be affected? 

Or maybe perhaps we add the feature to disableWindows. But it is counterintuitive in nature....considering all features in helm values are addition add-ons, rather than removals...

Dilemma here.

mauriciopoppe: I think we can make it `enableWindows: false` by default but we'd need to bump the Helm Chart to v2 because there might be people who already rely on this manifest being installed by default. The more we wait the harder it's going to make this switch.

@andyzhangx are you installing LVP through the Helm chart?

jennwah: @mauriciopoppe Alright lets change to `enableWindows: false` by default. I have changed the necessary and also added into documentation as well.

mauriciopoppe (commented):
Looking better, a few comments:

- We have chart examples in https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/tree/master/helm/examples I think you need to run https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/hack/update-generated.sh as well, please try to create a new commit for the generated changes.
- We also need to bump the version of the chart, this is done in https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/be7c6372d53dc7aa248ebb6738d71e553e1d9dbc/helm/provisioner/Chart.yaml#L5. 
- This change https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/386 is also ongoing which will create a conflict in the line above, I believe it's a breaking change so I think we could go for the safe route and bump it to v2

jennwah:
Let me know if i need to squash into single commit at the end, but I have ran the generated scripts for generated examples and also bump it to v2 for helm chart.

helm/provisioner/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: local-static-provisioner
 description: Helm chart for the SIG Storage Local Volume Static Provisioner.
 type: application
-version: 1.0.0
+version: 2.0.0
```
mauriciopoppe, line 5: Setting `enableWindows: false` works but we'll make the v2.0.0 bump after a few more commits, could you please undo this line?

mauriciopoppe:
We flattened a few values in #393 so please update your copy adding the `enableWindows` flat as a top level value

helm/provisioner/Chart.yaml
jennwah: alright reverted back to v1

mauriciopoppe:
Thanks a lot for your patience @jennwah!

/lgtm
/approve

PR #393, opened 2023-07-21
Flatten helm keys common, daemonset. Add CHANGELOG file for the helm chart

mauriciopoppe:
**What type of PR is this?**
>
/kind cleanup
/kind feature

For the v2 version of the chart let's flatten some of the nested properties.

Testing:
- After the refactor I run `./hack/verify-generated.sh` and there were no changes in the generated manifests.

**Release note**:
```release-note
Helm Chart values flattened when possible, please check the CHANGELOG.md file inside the helm/ directory
```

/cc @stevehipwell
/assign @msau42


helm/README.md
```diff
@@ -186,48 +186,47 @@ information.
...
+| fullnameOverride                        | Override default releasename                                                                          | str      | `""`                                                       |
+| rbac.create                             | if `true`, create and use RBAC resources                                                              | bool     | `true`                                                     |
+| serviceAccount.create                   | if `true`, create serviceaccount in .Release.Namespace                                                | bool     | `true`                                                     |
+| serviceAccount.name                     | if set serviceaccount if the given name will be created                                               | str      | `""`                                                       |
+| useAlphaAPI                             | If running against pre-1.10 k8s version, the `useAlphaAPI` flag must be enabled.                      | bool     | `false`                                                    |
```
msau42, line 196: Maybe since we're going to bump a major version, we can remove this as well.

mauriciopoppe: I removed the key from the helm chart too, pending is the removal of the go code that uses this field.

PR #401, opened 2023-08-03
feat(helm): Added support for additional volumes

stevehipwell:
**What type of PR is this?**
/kind feature

This PR adds support for providing volumes to init containers and the main container. A specific use case would be adding a script to an init container via a ConfigMap. This is done via the addition of `additionalVolume`s & `additionalVolumeMounts` inputs which allow for the addition of pod volumes and also support mounting extra volumes to the default container (init container volume mounts are part of the `initContainers` value).

It also removes the `additionalHostPathVolumes` value which wasn't ever used to mount a volume to the main container as it was only used in the templates for `volumes` and not `volumeMounts`.

This PR supercedes #386.

Fixes #

**Release note**:
```release-note

```

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -6,17 +6,18 @@ local-volume-provisioner in your Kubernetes with `helm install` directly.
 
 ## Table of Contents
 
-- [Install Helm](#install-helm)
-- [Custom your deployment with values file](#custom-your-deployment-with-values-file)
-- [Install local-volume-provisioner](#install-local-volume-provisioner)
-- [helm version < v3.0.0](#helm-version-lt-v3.0.0)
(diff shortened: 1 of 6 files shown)


mauriciopoppe (commented):
Looks good, do we need to change the example values and their generated output too? Ref https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/hack/verify-generated.sh

stevehipwell:
@mauriciopoppe the removed value wasn't used anywhere so we should be good.

mauriciopoppe (commented):
Could you please add an example and its generated output setting both `additionalVolumes` and `additionalVolumeMounts`? Those types of snapshot tests have helped a lot in the past when making refactors.

mauriciopoppe:
It's available now

PR #408, opened 2023-09-28
Add Cleanup Controller Deployment to Helm Chart

ripa1993:
**What type of PR is this?**

/kind feature

Add configurable Deployment for Cleanup Controller in Helm Chart, starting from the samples in `deployment/kubernetes/example/node-cleanup-controller`.

While testing storage-local-static-provisioner in a setup on EKS with EC2 instances with nvme disks, I noticed that the controller is a requirement in order to be able to have a proper scale down of the node groups, otherwise the "old" PVs are  kept even if the nodes are no longer available. I also noticed that for cluster-autoscaler to work properly with this controller, I need to configure a "fake" out-of-tree storage class provisioner to properly trigger the scale up of the the ASG when a Pod needs to bind a PVC

Fixes #

**Release note**:
```release-note

```


zchenyu:
Any progress on this PR?

It would be great if we could actually include this into the provisioner chart: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/tree/master/helm/provisioner
This seems like something that we should turn on by default.

PR #410, opened 2023-10-16
fix: do not try to get node for non-local PVs

prasadkatti:
Remote source PVs (like AWS EBS) do not have a node associated with them.
This commit skips the node lookup for such PVs.

Otherwise, the controller logs the following error for each remote PV:

```bash
error getting node attached to pv: &PersistentVolume{...}
```

**What type of PR is this?**

/kind bug

This commit skips the node lookup for non-local PVs, thereby preventing unnecessary errors in the logs.

Fixes #

**Release note**:
```release-note
NONE
```


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

PR #413, opened 2023-10-19
feat: add allowedTopologies for storage class in helm

AndrewChubatiuk:
**What type of PR is this?**
> /kind feature

Added support of allowedTopologies to allow mounting local volume to a pod only on nodes filtered by certain labels
Added ability to add additional rbac rules for a case when K8S API is triggered by init container

`allowedTopologies` variable can be as a dict (and  it will be converted to yaml) or as a yaml string template (useful to generate allowedTopologies from nodeAffinity)

**Release note**:
```release-note
added support of allowedTopologies to storageclass in helm
```

### DIFF
--- a/helm/provisioner/values.yaml
+++ b/helm/provisioner/values.yaml
@@ -5,6 +5,8 @@ fullnameOverride: ""
 rbac:
   # rbac.create: `true` if rbac resources should be created
   create: true
+  # list of additional rbac rules that may be needed by init containers
+  extraRules: []
 
 # Defines whether to generate a serviceAccount
 serviceAccount:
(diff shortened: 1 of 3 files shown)


andyzhangx (commented):
/ok-to-test

helm/provisioner/values.yaml
```diff
@@ -67,6 +69,8 @@ classes:
     fsType: ext4
     # File name pattern to discover. By default, discover all file names.
     namePattern: "*"
+    # Restrict topology of provisioned volumes to specific labels
+    allowedTopologies:
```
andyzhangx, line 73: does this empty setting work? I suppose it should write like this `allowedTopologies: {}`

andyzhangx (approved):
/lgtm

PR #417, opened 2023-11-10
feat: added hostPID capability for linux daemonset

jamesalford:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind flake

/kind feature

Adds the ability to add hostPID to the container spec for linux daemonset. This is used to execute scripts in init containers that modify the host storage

Fixes #

**Release note**:
```release-note

```

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -231,6 +231,7 @@ provisioner chart and their default values.
 | serviceMonitor.namespace                | The namespace Prometheus servicemonitor will be installed                                                                      | str      | `.Release.Namespace`                                          |
 | serviceMonitor.additionalLabels         | Additional labels for the servicemonitor                                                                                       | map      | `-`                                                           |
 | serviceMonitor.relabelings              | Additional metrics relabel_config                                                                                              | lists    | `-`                                                           |
+| hostPID                                 | Add `hostPID: true` to the container spec for linux daemonset.                                                                 | bool     | `false`                                                       |
 
 Note: `classes` is a list of objects, you can specify one or more classes.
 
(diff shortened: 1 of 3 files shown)


andyzhangx (commented):
/ok-to-test

helm/provisioner/templates/daemonset_linux.yaml
```diff
@@ -28,6 +28,9 @@ spec:
 {{ .Values.podAnnotations | toYaml | trim | indent 8 }}
 {{- end }}
     spec:
+{{- if .Values.hostPID }}
+      hostPID: true
```
mauriciopoppe, line 32: Nit: Maybe this should be:

```
hostPID: {{ .Values.hostPID }}
```

The default value in the reference is `false` https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#podspec-v1-core, also the value set from the chart is `false` so I think the above would work.

helm/README.md
```diff
@@ -231,6 +231,7 @@ provisioner chart and their default values.
 | serviceMonitor.namespace                | The namespace Prometheus servicemonitor will be installed                                                                      | str      | `.Release.Namespace`                                          |
 | serviceMonitor.additionalLabels         | Additional labels for the servicemonitor                                                                                       | map      | `-`                                                           |
 | serviceMonitor.relabelings              | Additional metrics relabel_config                                                                                              | lists    | `-`                                                           |
+| hostPID                                 | Add `hostPID: true` to the container spec for linux daemonset.                                                                 | bool     | `false`                                                       |
```
mauriciopoppe, line 234: If you consider the comment below then the docs should be updated too.

PR #435, opened 2024-03-06
Possibilty to add extra options to blkdiscard command

smallo:
In AWS EC2 instances `blkdiscard` is not working when cleaning a block device for Aerospike as per https://support.aerospike.com/s/article/How-to-decide-if-blkdiscard-is-preferable-to-dd

So, to do a fast clean up one might want to execute something like:
```
blkdiscard -z --length 8MiB /mnt/dev/some_device
```

By adding the possibility of sending extra options to `blkdiscard.sh`, executing the command above will be possible by configuring:
```
    local-ssd:
       hostDir: /mnt/disks
       mountDir:  /mnt/disks
       blockCleanerCommand:
         - "blkdiscard.sh"
         - -z --length 8MiB
       volumeMode: Block
```

### DIFF
--- a/deployment/docker/scripts/blkdiscard.sh
+++ b/deployment/docker/scripts/blkdiscard.sh
@@ -16,7 +16,7 @@
 
 
 # Usage:
-# $ blkdiscard.sh
+# $ blkdiscard.sh <blkoptions extra options>
 
 # Import common functions.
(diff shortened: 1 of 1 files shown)


andyzhangx:
@smallo can you sign the cla first? thx

PR #443, opened 2024-04-24
Fix: [node-cleanup] Check node label to know if pv is linked to a deleted node

Fricounet:
**What type of PR is this?**

/kind bug

The current implementation of the node cleanup controller has a bug where if the `kubernetes.io/hostname` in the affinity of the PV is not the same as the `node.Name`, all the local PV and PVC will be considered as being linked to a deleted node and be deleted. We have this situation on AWS clusters where the label is `kubernetes.io/hostname: ip-10-10-10-10` but the node name is `ip-10-10-10-10.ec2.internal`. 
This PR changes how the controller look for the node to now rely on the label on the node rather than  the node name. 

Fixes #

**Release note**:
```release-note

```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -33,9 +33,10 @@ import (
 	"sigs.k8s.io/yaml"
 
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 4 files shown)


Fricounet:
@andyzhangx do you think you could take a look at this PR when you have some time?

pkg/common/common.go
```diff
@@ -488,12 +489,17 @@ func GetVolumeMode(volUtil util.VolumeUtil, fullPath string) (v1.PersistentVolum
 }
 
 // NodeExists checks to see if a Node exists in the Indexer of a NodeLister.
-func NodeExists(nodeLister corelisters.NodeLister, nodeName string) (bool, error) {
-	_, err := nodeLister.Get(nodeName)
-	if errors.IsNotFound(err) {
-		return false, nil
+// It uses the well known label `kubernetes.io/hostname` to find the Node.
+func NodeExists(nodeLister corelisters.NodeLister, nodeLabelValue string) (bool, error) {
+	req, err := labels.NewRequirement(NodeLabelKey, selection.Equals, []string{nodeLabelValue})
```
andyzhangx, line 494: can you use the original method `_, err := nodeLister.Get(nodeName)` and if not found, use label as fallback, that would be safer, thx

Fricounet: good point. I've modified the code to keep the original get

andyzhangx (approved):
/lgtm

PR #466, opened 2024-11-18
helmchart: allow parameterize storageClass.provisioner

### LINKED ISSUE #288: Allow autoscaling based on local static provisioned volumes
**Is your feature request related to a problem? Please describe.**
To summarise: in an autoscaling scenario the autoscaler cannot scale based in volumes provisioned by the local static provisioner, as the usage of `provisioner: "kubernetes.io/no-provisioner"` prevents the autoscaling from been triggered due to how the [scheduler binder](https://github.com/kubernetes/kubernetes/blob/473af0b8d1951b5c362df60925bb8d9fb007fab9/pkg/controller/volume/scheduling/scheduler_binder.go#L868) works

The full problem is described in [the autoscaler repository](https://github.com/kubernetes/autoscaler/issues/1658#issuecomment-1036205889) by myself. But after thinking about it, I think it somehow makes more sense that it is solved in the provisioner side.

First of all, I have no insight on how CSI drivers work, but a solution to this issue could be that, optionally, the local provisioner could deploy a simple csidriver, that somehow allowed the PVCs to be bound to the PV created by the static provisioner. My guess is that the driver should "inform" the PVC about the name of the newly created PV, but this is just a guess.

Introducing another keyword instead of `provisioner: "kubernetes.io/no-provisioner"` that could be used to indicate to the scheduler binder that the volume will exist, even although it does not exist now (like it does with dynamic provisioners). This would need to be changed in the Kubernetes scheduler binder first, and then introduced in the local static provisioner.

**Additional context**
Full context in the mentioned comment and some previous comments in that thread: https://github.com/kubernetes/autoscaler/issues/1658#issuecomment-1036205889

lizzzcai:
**What type of PR is this?**
/kind feature

Allow parameterize the storageClass.provisioner in the Helm chart.
When I tested it in our environment, we faced a scaling issue similar to the issue mentioned below. The cluster autoscaler cannot spin up a new node for a pod using the volume provisioned by this storage class. The problem was resolved after overwriting the provionser from `kubernetes.io/no-provisioner` to another value (workaround mentioned [here](https://github.com/kubernetes/autoscaler/issues/1658#issuecomment-1036205889)).

Fixes #288

Manually tested and working fine:
```sh
❯ helm template ./helm/provisioner --values custom-values.yaml --dry-run | grep -i storageclass -A 10 | grep provisioner
# Source: local-static-provisioner/templates/storageclass.yaml
    helm.sh/chart: local-static-provisioner-2.0.0
    app.kubernetes.io/name: local-static-provisioner
provisioner: my-provisioner
```

**Release note**:
```release-note
Allow parameterize the storageClass.provisioner in the Helm chart.
```


andyzhangx (approved):
/lgtm

PR #501, opened 2025-06-17
Support for removing start up taint

saidjawad:
**What type of PR is this?**
/kind feature

This PR adds support for removing [node startup taints](https://github.com/aws/karpenter-provider-aws/issues/628) with an approach similar to [Cilium](https://docs.cilium.io/en/stable/gettingstarted/k8s-install-default/) and [aws ebs csi driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1232). 

When nodes start up or get provisioned, there is certain delay till the local-static-provisioner becomes ready. This could cause race condition with the scheduler, because other workload pods with a local PVC request might get scheduled on the node and stuck waiting for their volumes be provisioned. 

This PR assumes that the node provisioner already adds some start up taints, and when the local-static-provisioner becomes ready, removes them. 

We do it only once at the node starts up.

Fixes #

**Release note**:
```release-note

```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/common"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/deleter"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/discovery"
+	nodetaint "sigs.k8s.io/sig-storage-local-static-provisioner/pkg/node-taint"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/populator"
 	"sigs.k8s.io/sig-storage-local-static-provisioner/pkg/util"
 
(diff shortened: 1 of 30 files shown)


saidjawad:
@andyzhangx Seems that tests are passing. Is there anything else that need to be done?

pkg/controller/controller.go
```diff
@@ -174,6 +178,9 @@ func StartLocalController(signal *signal, client *kubernetes.Clientset, ptable d
 		default:
 			deleter.DeletePVs()
 			discoverer.DiscoverLocalVolumes()
+			if !nodeTaintRemover.ShouldRemoveTaint() && discoverer.Readyz.Check(nil) == nil {
```
andyzhangx, line 181: refer to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1588, why not use a backoff try if taint removal fails?

andyzhangx (approved):
/lgtm

PR #528, opened 2025-12-02
Update comments to drop very old version

kylepl:
**What type of PR is this?**
/kind documentation

* Remove another references to feature flags missed in https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/commit/827dbf564e8f6d7ae37a078b52fb6990c298b949
* Remove a few references to version 1.9+, which, assuming it is k8s (or even for this provisioner), they are extremely old.

I was not 100% which version 1.9+ applies too, but seems extremely old either way.


andyzhangx (approved):
/ok-to-test
/lgtm

PR #531, opened 2025-12-12
Helm: support accessMode and update image tag

dud225:
**What type of PR is this?**
/kind bug

This brings support of the `accessMode` setting in storageClassMap and update the image tag.


andyzhangx (approved):
/lgtm

